PR #256, opened 2015-04-30
Adding styles to the shapes of bpmn

fmcejudo:
Hello:

I have made a bunch of changes to support styles at the shapes, just including class attribute to the svg tags and id, which corresponds with the bpmn definition.

The id allows as well to work with events, what have let me drill down into  sub-processes.

Thank you. 

Cheers

### DIFF
--- a/bpmn-js
+++ b/bpmn-js
@@ -1,0 +1,1 @@
+/home/francisco/workspaces/thirdparties/bpmn-js
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -1,17 +1,19 @@
 'use strict';
 
 var inherits = require('inherits'),
-    isArray = require('lodash/lang/isArray'),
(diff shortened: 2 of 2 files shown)


nikku:
Could you elaborate a bit more on your "use case", i.e. the stuff you want to achieve?

fmcejudo:
what I wanted to get was a process diagram which represents no just the process but as well the status of an instance of that process, such as the past stage, the current and the future. 

Also it lets navigate the process and their call activities, and going back to the parent if it was the case.

Those styles depend on the user to complete the css stylesheet with the name of the shapes.

nikku:
So you basically would like to have additional element attributes to add better styling via CSS?

It would be cool if you could provide an actual example that show cases how this is useful. We'd need something like this as part of our [examples](https://github.com/bpmn-io/bpmn-js-examples) anyway before we can get your PR merged.

nikku:
Closing this due to inactivity.

Issue #231, opened 2015-03-25
Sequence flows inside collapsed sub processes are being rendered


nikku:
See [forum thread](https://forum.bpmn.io/t/81) that attaches an example diagram, too.

PR #257, opened 2015-05-01
It splits subprocesses from call activities

fmcejudo:
Hello:

I found a problem representing my subprocesses, they didn't show the inner components as they should.

I checked some subprocesses you are testing and I imported them, and they work, but if I got and I modified them with external IDEs, and I re-imported to bpmn.io demo, they were not well represented.

I have modified a little bit how the context behaves, just to show the inner components, and with them I had to modify as well the tests. 

I have probed both the ones you are testing and the ones I had from Eclipse with Activity, and they are represented as I expected.

After doing the changes:
withchanges

As currently it  draws the subprocess:

ascurrentlyis

I hope it could work for you.

Thank you.

Cheers.

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -1,17 +1,17 @@
 'use strict';
 
 var inherits = require('inherits'),
-    isArray = require('lodash/lang/isArray'),
-    isObject = require('lodash/lang/isObject'),
-    assign = require('lodash/object/assign'),
-    forEach = require('lodash/collection/forEach'),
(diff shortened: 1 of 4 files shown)


nikku:
Looks like Activiti/Eclipse do not properly expand sub processes. Our tool currently renders connections inside collapsed sub processes (bug #231). Thus the picture.

We have [test cases](https://github.com/bpmn-io/bpmn-js/blob/master/test/spec/import/elements/CollapsedSpec.js#L17) that verify we correctly import collapsed sub processes.

We cannot merge this PR. It does not fix any known issue with our tool.

PR #270, opened 2015-05-08
fix(context-pad): fix context pad positioning in special case

mhariri:
context pad was misplaced when the canvas is placed with a left offset within the parent

### DIFF
--- a/lib/features/context-pad/ContextPadProvider.js
+++ b/lib/features/context-pad/ContextPadProvider.js
@@ -77,9 +77,10 @@ ContextPadProvider.prototype.getContextPadEntries = function(element) {
         padRect = pad.getBoundingClientRect();
 
     var top = padRect.top - diagramRect.top;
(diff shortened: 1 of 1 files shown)


nikku:
Could you elaborate a bit more on your "special case"? I.e. which usage scenario does require the fix?

mhariri:
To reproduce:
1. Go to http://demo.bpmn.io/new
2. Using the chrome's html editing feature, create a table with two columns, and put the bjs-container inside the second column of that table.
3. Click on the diagram's "tool" button

You will see that the tools menu appears far away from the diagram element, as shown in the screenshot. The patch fixes this issue.

mhariri:
screen shot 2015-05-08 at 18 15 38

nikku:
I followed your steps and could not reproduce your issue on [demo.bpmn.io](http://demo.bpmn.io).

PR #359, opened 2015-09-09
Added functionality to copy properties while replaceing element

korbinianHoerfurter:
This is an additional functionality for copiing all properties of an bpmn-element to another while replaceing ist (i.e. Change task type)

The code checks whether the old properties fit to the new element.


pedesen:
Closed due to inactivity

pedesen:
I reopened the ticket again, because it is a nice feature that we should implement ourselves.

korbinianHoerfurter:
Hi,
the problem was, that the tests failed also before I added the new functionality (directly after the fork).

nikku:
Closing this now. We will work on the functionality via #415.

PR #375, opened 2015-10-05
feat(ids): unclaim old business object id

polenz:
Required for bpmn-io/bpmn-js-properties-panel#23


nikku:
Merged via 188487ebdc29c3d56c7e45930e4b5ba22a337011

PR #409, opened 2015-11-10
fix(replace): do NOT show current sequence flow type for OR-gateways

### LINKED ISSUE #410: Default sequence flow can be replaced with both 'sequence flow' and 'default sequence flow' for OR-gateways
or-gateway

adbre:
Closes #410


pedesen:
Merged via https://github.com/bpmn-io/bpmn-js/commit/0a33d7c2b497a540a09d85ff63aa10357dbbc087

Thanks a lot for that pull request. It's very much appreciated!

PR #419, opened 2015-11-24
fix: handle empty model

### LINKED ISSUE #417: Error if there is no diagram to display
The library should indicate via an error, if there was nothing to display due to missing DI or import error.

Currently we do nothing to mitigate this issue and run into trouble as soon as the user wants to model something.

#### Example Diagram

```
<bpmn:definitions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:bpmn="http://www.omg.org/spec/BPMN/20100524/MODEL" xmlns:bpmndi="http://www.omg.org/spec/BPMN/20100524/DI" xmlns:dc="http://www.omg.org/spec/DD/20100524/DC" targetNamespace="http://bpmn.io/schema/bpmn" id="Definitions_1">
</bpmn:definitions>
```

#### Details

[`BpmnJS#importXML`](https://github.com/bpmn-io/bpmn-js/blob/master/lib/Viewer.js#L159) should indicate the error to the user as shown below:

```js
bpmnjs.importXML('no-diagram-xml.bpmn', function(err) {
   // err = Error{ no diagram to display }
});
```

dlw93:
Return an error message in case the XML does not contain any valid diagrams
Closes #417

### DIFF
--- a/lib/Viewer.js
+++ b/lib/Viewer.js
@@ -168,6 +168,11 @@ Viewer.prototype.importXML = function(xml, done) {
       err = checkValidationError(err);
       return done(err);
     }
+	
+    if(context.parseRoot.element.diagrams === undefined) {
+      err = new Error("No diagrams to display");
+      return done(err);
+    }
 
     var parseWarnings = context.warnings;
 
--- a/lib/Viewer.js
+++ b/lib/Viewer.js
@@ -169,8 +169,8 @@ Viewer.prototype.importXML = function(xml, done) {
       return done(err);
     }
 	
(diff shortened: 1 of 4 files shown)


nikku:
Could you add test cases to verify your solution works?

nikku:
Thanks for the updates. I will look into your PR this week.

nikku:
Merged with changes via 6e2ec62b00793e6a24629d1bf90186d614f50681.

Stuff I adjusted:

* detect and handle missing diagram in BpmnTreeWalker
* squashed commits into one and adjusted commit message

PR #424, opened 2015-12-02
feat(ContextPad): Do not show delete action when not allowed

adbre:
See bpmn-io/diagram-js#131


nikku:
Merged via 9322a6a1e65606f9a29e9970a74c050c4a4ba4fc with minor changes (mostly coding style fixes).

I removed the mocking from the test cases, too as there was an easy way to work without it.

Thanks for your contribution!

PR #426, opened 2015-12-03
feat(BpmnReplace): Let rules disallow replacement operations

adbre:
From discussion at https://forum.bpmn.io/t/rule-for-when-deleting-a-element-is-permitted/387/9


nikku:
Merged via 93d4bade22ce403d4f4f49fa930f9e8999980a2a.

Thanks for this great PR.

PR #434, opened 2015-12-16
fix(draw): add plus sign (+) to call activities

### LINKED ISSUE #403: A call activity is always shown with a "+" sign
We need to update our [`BpmnRenderer`](https://github.com/bpmn-io/bpmn-js/blob/master/lib/draw/BpmnRenderer.js) to always show the plus sign (__+__) for call activities.

The collapsed sub-process as shown below already has the marker (left). The call activity should get it, too.

image

Test diagram for which that should work:

```xml
<bpmn:definitions xmlns:bpmn="http://www.omg.org/spec/BPMN/20100524/MODEL" xmlns:bpmndi="http://www.omg.org/spec/BPMN/20100524/DI" xmlns:di="http://www.omg.org/spec/DD/20100524/DI" xmlns:dc="http://www.omg.org/spec/DD/20100524/DC" id="Definitions_1" targetNamespace="http://bpmn.io/schema/bpmn">
  <bpmn:process id="Process_1" isExecutable="false">
    <bpmn:subProcess id="SubProcess_02yxyt7" name="A" />
    <bpmn:callActivity id="CallActivity_1acw786" name="B" />
  </bpmn:process>
    <bpmndi:BPMNPlane id="BPMNPlane_1" bpmnElement="Process_1">
      <bpmndi:BPMNShape id="SubProcess_02yxyt7_di" bpmnElement="SubProcess_02yxyt7">
        <dc:Bounds x="245" y="130" width="100" height="80" />
      </bpmndi:BPMNShape>
      <bpmndi:BPMNShape id="CallActivity_1acw786_di" bpmnElement="CallActivity_1acw786">
        <dc:Bounds x="399" y="130" width="100" height="80" />
      </bpmndi:BPMNShape>
    </bpmndi:BPMNPlane>
  </bpmndi:BPMNDiagram>
</bpmn:definitions>
```

hoferch91:
Closes #403

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -854,7 +854,7 @@ function BpmnRenderer(eventBus, styles, pathMap, priority) {
       return outer;
     },
     'bpmn:CallActivity': function(p, element) {
-      return renderer('bpmn:Task')(p, element, {
+      return renderer('bpmn:SubProcess')(p, element, {
         strokeWidth: 5
       });
     },


nikku:
Could you add a test case that verifies that the plus is actually rendered for Call Activities?

An option would be to see if it is present in the resulting SVG element.

PR #439, opened 2015-12-23
fix(draw): add plus sign (+) to call activities #434

hoferch91:


nikku:
Thanks for your PR. I merged it via 45f01d995a3f3165c25840104af3bf582705df5c with one change:

I added a `data-marker` attribute to the marker elements, allowing us to test whether a marker exists much more easily.

Issue #466, opened 2016-02-03
Localization Support (Different Language)


havid0707:
There does not exists one way to localize the modeler UI without directly modifying the source code at this moment. 

Please help to make it easy to plug-in additional languages.

nikku:
We accept pull-requests for that.

See [this forum topic](https://forum.bpmn.io/t/modeler-localization/255) for a discussion.

nikku:
Closed via bpmn-io/bpmn-js#491.

PR #475, opened 2016-02-25
feat(i18n): Localization with the i18next library.

davcs86:
Adds multi-language support to bpmn-js using the [i18next](http://i18next.com/) library (looking to solve #466)

It exposes an i18next instance through di.

```
// load the translation
modeler.get('i18n').addResourceBundle('es-MX', 'translation', {'Remove': 'Eliminar'});
// change the language
modeler.get('i18n').changeLanguage('es-MX');
```

Notes:

- i18next API details [here](http://i18next.com/docs/api/).
- Tests [here](https://github.com/davcs86/bpmn-js/blob/i18n/test/spec/features/i18n/I18nSpec.js#L9).
- Default settings [here](https://github.com/davcs86/bpmn-js/blob/i18n/lib/features/i18n/DefaultOptions.js#L1).
- Translation function in modules: `i18n.t('key')`, after inject the dependence `i18n`, here an [example](https://github.com/davcs86/bpmn-js/blob/i18n/lib/features/palette/PaletteProvider.js#L8).


nikku:
Thanks for your PR.

For internationalization we would like to keep the overall API surface as small as possible. That means no additional libraries for English (the default language).

The best way to achieve this is to make the internationalization pluggable. Provide a really slim default implementation and let users decide whether they want to plug-in something else (i.e. `i18next`). That naturally increases the bundle size. 

I propose the following:

*   Instead of using `i18next` provide a simple `translate` helper.
*   It exposes a simple API:
```javascript
function translate(message, additionalArgs....) {
  return ....; // some simple default implementation
}
```
*   It should be possible to invoke the helper in a way that provides all the required data:

```javascript
   if (typeof title !== 'string') {
     options = title;
-    title = 'Append ' + type.replace(/^bpmn\:/, '');
+    title = translate('Append {0}', type.replace(/^bpmn\:/, ''));
   }
```
* There is a variant of translate that allows translation with another library, i.e. `i18next`:

```javascript
function translate(message, additionalArgs....) {
  return i18next.translate(...);
}
```

nikku:
We are happy to merge the internationalization changes if you adjust your implementation accordingly.

davcs86:
I changed the approach as you suggested, by replacing `i18n` with a helper with a basic string interpolation as the [default functionality](https://github.com/davcs86/bpmn-js/commit/7188239dedd81d962c371ffc7dbd20fd19fb05b4#diff-1c3589896e61c44683e8fd8498768b8cR454).

With this, the user only has to override the function `t` of the helper, e.g.

```js
//(...)
var modeler = new Modeler({ container: container });
modeler.get('translate').t = customTranslatorFunc;
// and to update the palette (or any other control in the future), notify the language change with
modeler.get('translate').applyLanguage();
// (...)
// where
function customTranslatorFunc(str, args){
    // if using i18next
    return i18next.t(str, args);
}
```

santosguilherme:
How i use translate feature?

davcs86:
You can create an inherated module of `bpmn-js/lib/features/translate/TranslateProvider` and use your favorite localization library. 

Example with i18next: [bpmn-js-i18next](https://github.com/davcs86/bpmn-js-i18next)

santosguilherme:
I need extend the TranslateProvider?

davcs86:
Yes, A _quick and dirty_ option is to override the function `t` like mentioned in a previous comment. Something like this,

```js
modeler.get('translate').t = function(string, additionalArgs) {
    return yourTranslationFunc(string, additionalArgs);
};
modeler.get('translate').applyLanguage();
```

but in this way, you have to override the function each time a diagram is created or imported.

santosguilherme:
I tried this, but I received an error on console, something like "No provider to the module 'translate'."

My code:
`new BpmnJS({container: '#js-canvas'}).get('translate')`

Algo errado?

davcs86:
Yes, you have to use the PR branch since this is still not merged in the master version.

santosguilherme:
There ok, any prevision for when PR branch will be merged?

Thanks for help @davcs86.

davcs86:
I hope soon, it doesn't depend on me.

Meanwhile, you can use test it with,

**bower**

`bower install --save davcs86/bpmn-js#i18n`

or 

**npm**

`npm install --save davcs86/bpmn-js#i18n`

havid0707:
HI all, may i ask this can be merged now?

nikku:
We are going to look into this PR again and merge it soon. 

The next bpmn-js release will ship with localization support. @davcs86 Thanks again for contributing.

davcs86:
Sure @nikku, feel free to let me know any comment, question or ask a change that could help to merge this.

nikku:
Closed via #491 with some small adjustments.

davcs86:
Hi @nikku, 

I'm checking out the changes. And I have a doubt, if

```
module.exports = {
  translate: [ 'value', require('./Translate') ]
};
```

implies I can't inject dependencies since it only will return the function as is. So, how can I override it in a modular way? I know you want to make this environment-agnostic but I feel this some inflexible.

At least, I would like to know how can I override `translate` function from another module.

nikku:
Can be easily overridden (see [this test case](https://github.com/bpmn-io/bpmn-js/commit/49173abdade56a598b4bdbb3921090ed98f8a5d3#diff-c6b77c29a796ec90a166e980872981daR17)).

You could expose a factory (that allows injection), too:

```javascript
module.exports = {
  translate: [ 'factory', function(eventBus, i18n) {
    return function translate(...) {};
  } ]
};
```

davcs86:
wow! very simple, thanks.

santosguilherme:
How to override the translate function instead of the module translate?

nikku:
[See above](https://github.com/bpmn-io/bpmn-js/pull/475#issuecomment-196411836). This is how you do it.

santosguilherme:
Okay, but how will override from an angularjs controller's without browserify?

nikku:
```javascript

function customTranslate(template, replacements) {
  // TODO: do actual translation
}

var bpmnModeler = new BpmnModeler({
  additionalModules: [
    {
      translate: [ 'value', customTranslate ]
    }
  ]
});
```

santosguilherme:
Thanks @nikku

santosguilherme:
Is there a file with all possible keys  to translate?

davcs86:
So far, I've used these. I'm not sure if they added more.

```json
{
    'Append {type}': 'Append {type}',
    'Add Lane above': 'Add Lane above',
    'Divide into two Lanes': 'Divide into two Lanes',
    'Divide into three Lanes': 'Divide into three Lanes',
    'Add Lane below': 'Add Lane below',
    'Append compensation activity': 'Append compensation activity',
    'Change type': 'Change type',
    'Connect using Association': 'Connect using Association',
    'Connect using Sequence/MessageFlow or Association': 'Connect using Sequence/MessageFlow or Association',
    'Connect using DataInputAssociation': 'Connect using DataInputAssociation',
    'Remove': 'Remove',
    'no parent for {element} in {parent}': 'no parent for {element} in {parent}',
    'no shape type specified':'no shape type specified',
    'flow elements must be children of pools/participants': 'flow elements must be children of pools/participants',
    'out of bounds release': 'out of bounds release',
    'more than {count} child lanes': 'more than {count} child lanes',
    'element required': 'element required',
    'Create {type}': 'Create {type}',
    'Activate the hand tool': 'Activate the hand tool',
    'Activate the lasso tool': 'Activate the lasso tool',
    'Activate the create/remove space tool': 'Activate the create/remove space tool',
    'Create expanded SubProcess': 'Create expanded SubProcess',
    'Create Pool/Participant': 'Create Pool/Participant',
    'Parallel Multi Instance': 'Parallel Multi Instance',
    'Sequential Multi Instance': 'Sequential Multi Instance',
    'Loop': 'Loop',
    'Ad-hoc': 'Ad-hoc',
    'element {element} referenced by {referenced}#{property} not yet drawn': 'element {element} referenced by {referenced}#{property} not yet drawn',
    'unknown di {di} for element {semantic}': 'unknown di {di} for element {semantic}',
    'missing {semantic}#attachedToRef': 'missing {semantic}#attachedToRef',
    '{semantic}#{side} Ref not specified': '{semantic}#{side} Ref not specified',
    'already rendered {element}': 'already rendered {element}',
    'failed to import {element}': 'failed to import {element}',
    'multiple DI elements defined for {element}': 'multiple DI elements defined for {element}',
    'no bpmnElement referenced in {element}': 'no bpmnElement referenced in {element}',
    'diagram not part of bpmn:Definitions': 'diagram not part of bpmn:Definitions',
    'no diagram to display': 'no diagram to display',
    'no plane for {element}': 'no plane for {element}',
    'no process or collaboration to display': 'no process or collaboration to display',
    'correcting missing bpmnElement on {plane} to {rootElement}': 'correcting missing bpmnElement on {plane} to {rootElement}',
    'unsupported bpmnElement for {plane}: {rootElement}': 'unsupported bpmnElement for {plane}: {rootElement}',
    'unrecognized flowElement {element} in context {context}': 'unrecognized flowElement {element} in context {context}'
}
```

nikku:
I added #497 to collect and provide a translation template (as shown above) as part of our releases. Cannot promise when exactly we are going to work on this.

santosguilherme:
Ok, thanks @nikku

PR #582, opened 2016-06-30
fix(replace):ComplexGateway can have default sequence flows

adbre:
The [BPMN 2.0 specification](http://www.omg.org/spec/BPMN/2.0/) states "a Sequence Flow that has an
Exclusive, Inclusive or Complex Gateway or an Activity as its source
can also be defined with as default."

This commit will honor that statement by allowing sequence flows from
Complex Gateway to be defined as default.


nikku:
Thanks for the PR.

PR #598, opened 2016-07-27
feat(modeling): Support disabled modeling

adbre:
Implements required changes introduced by bpmn-io/diagram-js#185


adbre:
Abandoning this PR, as a different approach is preferred.
See the [forum thread](https://forum.bpmn.io/t/changing-from-modeler-to-viewer-and-back-again/900/7?u=adbre) for info.

PR #614, opened 2016-09-22
fix(modelling) allow move of a seclection in different pools

### LINKED ISSUE #524: Impossible to move elements from different pools
Via [this forum topic](https://forum.camunda.org/t/dragging-task-boxes-in-different-pool-participants/518).

2905445186f3087bcaff8937e59e87b5539701a6

__Steps to reproduce__

* Open [`diagram.bpmn.txt`](https://github.com/bpmn-io/bpmn-js/files/340985/diagram.bpmn.txt)
* Select `Task_A`, `Task_B` and the connecting flow
* Try to move all selected elements

__Problem__

* The move operation is prevented by rules

__Expected behavior__

* It should be possible to move all elements as long as the primary selection (the element behind the cursor) does not change its parent.

__Hints___

* The [`Move` feature](https://github.com/bpmn-io/diagram-js/blob/master/lib/features/move/Move.js#L35) checks continuously during element moving whether the operation is allowed.

jocutajar:
The move was prevented by MessageFlow element being included in the moved selection.

I have eased the restrictions on MessageFlow canDrop rules somewhat to satisfy a new test.

    Fixes #524

### DIFF
--- a/lib/features/rules/BpmnRules.js
+++ b/lib/features/rules/BpmnRules.js
@@ -420,7 +420,10 @@ function canDrop(element, target, position) {
   }
 
   if (is(element, 'bpmn:MessageFlow')) {
-    return is(target, 'bpmn:Collaboration');
+    return is(target, 'bpmn:Collaboration')
+          // re #524: allow message flow move as part of a selection
+          || element.source.parent == target
+          || element.target.parent == target;
   }
 
   return false;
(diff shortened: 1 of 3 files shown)


pedesen:
Thanks! Merged via https://github.com/bpmn-io/bpmn-js/commit/5441b2e09e2da2b20ec8f764f2f7b18a1e64198e.

PR #656, opened 2017-01-27
feat(Viewer): allow to attach and detach the viewer from DOM

### LINKED ISSUE #655: Make the viewer attachable
If I create an instance of the viewer then it will be automatically attached to the DOM. It would be nice if the viewer could be instantiated first and then later attached to the DOM. This would be great when using the viewer in a React context were the DOM isn't always available.

This could be made configurable. So if I set the `container` option then it will be automatically attached. Otherwise the user must attach it by itself.

The same functionality already exists for the [properties panel](https://github.com/bpmn-io/bpmn-js-properties-panel#dynamic-attachdetach).

akkie:
Fixes #655

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -119,6 +119,8 @@ module.exports = function(grunt) {
 
   grunt.registerTask('test', [ 'karma:single' ]);
 
+  grunt.registerTask('lint', [ 'eslint:check' ]);
+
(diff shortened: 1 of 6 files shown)


nikku:
Merged via 6dc4af53969825ae0122bf06e3964d74adfa1cc3, thanks!

PR #664, opened 2017-02-18
fix(release): Stronger check for jQuery before unwraping.

juanmi108:
Other libraries exits that adds the `.get()` functionality to element nodes, in my case Mootools which behaves in a different fashion resulting in multiple errors.
 
Since the comment indicates that the condition is only for JQuery users, it should only fire if the lib is present.


juanmi108:
updated the check as suggested in https://github.com/bpmn-io/diagram-js/pull/207

PR #672, opened 2017-03-15
Patched code to enable rendering of data collection marker

lgbanuelos:
The support for data collections in `bpmn-js` is currently incomplete and/or broken. More specifically, the class `BpmnRenderer` does not add the corresponding marker to the data object shape when attribute `isCollection` is set to true.

To describe the workaround that I found, let us assume that we set up the modeler's palette with the following snippet (based on the custom elements example):

```
'create.data-object-collection': createAction(
    'bpmn:DataObjectReference', 'data-object', 'bpmn-icon-data-object',
            'Create data object collection', {isCollection: true}
)
```
With the above, I can create a data object element with the attribute `isCollection` set to true. I found that `BpmnRenderer` decides whether to add the collection marker to the data object shape or not in lines 144-148:

```
var semantic = getSemantic(element);

if (isCollection(semantic)) {
   renderDataItemCollection(parentGfx, element);
} 
```

Hence, I patched the same lines of code as follows:

```
var semantic = getSemantic(element);

if (isCollection(semantic) || isCollection(element)) {
   renderDataItemCollection(parentGfx, element);
} 
```

which forces the `BpmnRenderer` to add the marker in the corresponding data object shape. 

What I found is that `element` (i.e. the `Shape` associated with the BPMN data object) carries the attribute `isCollection`, which is fine. However, the test `isCollection` occurs not at the level of `element` but deeper (i.e. in child objects).

I think of two possible options:
1.  The element factory is setting the attribute `isCollection` at the wrong level, i.e. `isCollection` is currently attached to a `Shape` object, but should be attached to the shape's `businessObject`, or
2. The verification of the condition `isCollection` should be done on `element` as suggested by my patch.

One additional problem I found in `bpmn-js` is that the attribute `isCollection` is not serialized in the BPMN XML document associated with a model. That is why I still believe that the problem could be on the way the element factory attaches the attribute `isCollection`.

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -1443,7 +1443,7 @@ function BpmnRenderer(eventBus, styles, pathMap, canvas, priority) {
 
       var semantic = getSemantic(element);
 
-      if (isCollection(semantic) || isCollection(element)) {
+      if (isCollection(semantic)) {
         renderDataItemCollection(parentGfx, element);
       }
 
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -1443,7 +1443,7 @@ function BpmnRenderer(eventBus, styles, pathMap, canvas, priority) {
 
       var semantic = getSemantic(element);
 
-      if (isCollection(semantic)) {
+      if (isCollection(semantic) || isCollection(element)) {
         renderDataItemCollection(parentGfx, element);
       }
 
--- a/lib/features/modeling/behavior/CreateDataObjectBehavior.js
+++ b/lib/features/modeling/behavior/CreateDataObjectBehavior.js
@@ -23,8 +23,10 @@ function CreateDataObjectBehavior(eventBus, bpmnFactory, moddle) {
       // create a DataObject every time a DataObjectReference is created
       var dataObject = bpmnFactory.create('bpmn:DataObject');
 
-      if (shape.isCollection || shape.businessObject.isCollection)
+      if (shape.isCollection) {
+        shape.businessObject.isCollection = true;
         dataObject.isCollection = true;
+      }
 
       // set the reference to the DataObject
       shape.businessObject.dataObjectRef = dataObject;
--- a/lib/features/modeling/behavior/CreateDataObjectBehavior.js
+++ b/lib/features/modeling/behavior/CreateDataObjectBehavior.js
@@ -23,6 +23,9 @@ function CreateDataObjectBehavior(eventBus, bpmnFactory, moddle) {
       // create a DataObject every time a DataObjectReference is created
       var dataObject = bpmnFactory.create('bpmn:DataObject');
 
+      if (shape.isCollection || shape.businessObject.isCollection)
+        dataObject.isCollection = true;
+
       // set the reference to the DataObject
       shape.businessObject.dataObjectRef = dataObject;
     }
--- a/lib/import/BpmnTreeWalker.js
+++ b/lib/import/BpmnTreeWalker.js
@@ -390,6 +390,8 @@ function BpmnTreeWalker(handler, translate) {
       } else if (is(e, 'bpmn:DataStoreReference')) {
         handleDataElement(e, context);
       } else if (is(e, 'bpmn:DataObjectReference')) {
+        if (e.dataObjectRef && e.dataObjectRef.isCollection)
+          e.isCollection = true;
         handleDataElement(e, context);
       } else {
         logError(


lgbanuelos:
After digging a bit further, I found out that the problem is elsewhere. So, I reverted the changes I made to `BpmnRendered`.

What I found is that, according to the BPMN specification, the attribute `isCollection` belongs to a `bpmn:DataObject` element. However, the modeler uses the information on `bpmn:DataObjectReference` to decide whether to render the marker or not. In fact, the modeler's palette is configured to create a `bpmn:DataObjectReference` and the corresponding `bpmn:DataObject` is then created as a side effect. Hence, what I propose you is to keep a copy of the attribute `isCollection` in both `bpmn:DataObject` and `bpmn:DataObjectReference`.

To accomplish the above I have done two things:

1. I patched the file `CreateDataObjectBehavior.js` to copy the attribute into the `bpmn:DataObject` instance that is created by the modeler.
2. I patched the file `BpmnTreeWalker.js` to copy the attribute into the `bpmn:DataObjectReference` instance, as part of the deserialization of an XML file.

nikku:
After looking into your PR I'm still not 100% sure what you'd like to achieve.

* Be able to create a data object with `isCollection=true` semantics from the palette (https://github.com/bpmn-io/bpmn-js/issues/381)
* Render a Data Object correctly

The second one already works.

nikku:
Ok, digging a bit deeper, rendering of collection markers seems to be broken.

nikku:
So actually, [this line](https://github.com/bpmn-io/bpmn-js/blob/master/lib/draw/BpmnRenderer.js#L1831) is the offender that breaks proper collection marker rendering.

nikku:
Not sure if _Be able to create a data object with isCollection=true semantics from the palette (#381)_ is a concern for you. The rendering should be fixed now.

lgbanuelos:
I am just wondering if you change also fixes the problem with serialization to/deserialization from XML. Can you confirm?

nikku:
What exactly would you like to achieve regarding serialization, de-serialization? Is it https://github.com/bpmn-io/bpmn-js/issues/381?

lgbanuelos:
Assume we draw a BPMN model with bpmn-js modeler with a data object collection. With the current patch, the marker will be properly rendered. The problem is that if I want to save the model in XML, the attribute `isCollection` is not serialized as part of the `bpmn:DataObject` element. The reason behind this problem seems to be that, within the modeler, the attribute `isCollection` is attached to a `bpmn:DataObjectReference` and not to the corresponding `bpmn:DataObject`.

Now, assume that we have an XML file for a model with a data object collection, where the attribute `isCollection` is present in the `bpmn:DataObject` (as required by BPMN's spec). When such XML file is parsed, the attribute `isCollection` is never copied into the corresponding `bpmnDataObjectReference`. As a result of this, bpmn-js' modeler/viewer will not render the marker.

That is what I tried to do with my patch:

1. The attribute `isCollection` is copied to the `bpmn:DataObject` when it is created within `CreateDataObjectBehavior.js`   (this way the attribute gets serialized into the XML file)
2. The attribute `isCollection` is copied to the `bpmn:DataObjectReference` after parsing the XML file, when bpmn-js wires up the model within the `BpmnTreeWalker.js`    (this way the attribute is copied into `bpmn:DataObjectReference` and forces the rendering of the marker)

nikku:
All right, so what you'd like to do is creating a Data Object with `isCollection` semantics in the modeler. The round-trip (from XML -> to XML) was always working, that is why I wondered. Sorry for the confusion! :tada:

nikku:
So what your PR is doing regarding model creation is monkeypatching the `isCollection` property into both, the `DataObjectReference` (not allowed here) as well as the `DataObject` (right place). 

What you should do instead in your `PaletteProvider` is:

* Create `DataObjectReference` and wire it with a new `DataObject{ isCollection: true }`
* Pass it as `businessObject` via `options` to `ElementFactory#createShape` (similar to [this line](https://github.com/bpmn-io/bpmn-js/blob/master/lib/features/palette/PaletteProvider.js#L47)).

Additionally, we'd need to add the following change:

* [Fix the `CreateDataObjectBehavior`](https://github.com/bpmn-io/bpmn-js/blob/master/lib/features/modeling/behavior/CreateDataObjectBehavior.js#L21) to only create a data object if the data object reference does not have one attached yet

lgbanuelos:
I agree with you, having the object collection icon in the palette is not critical. I was also thinking about exploring the code and try to set up the pop-up menu that you suggest. What seems important to me is to ensure that the XML round-trip works fine. In my side, the round-trip seems to be broken (I will double check it later, no time today to do it though).

nikku:
Just verified against demo.bpmn.io that the roundtrip works. Closing this PR for now.

PR #678, opened 2017-04-17
feat(replace): add id to boundaryEvent attached eventDefinition

paed01:
Hi authors,

First of all: Your work with this repo is impressive.

Though, I have encountered a problem with orphan references when importing models with `bpmn-moddle`. I tracked it down to this repo since it is referenced from the Camunda Modeler.

When importing e.g. an `errorEventDefinition` the error reference is orphaned because there is no id to reference. This goes for other eventDefinitions that need references as well.

```js
const BpmnModdle = require('bpmn-moddle');
const moddle = new BpmnModdle();

const withoutIdDef = `<?xml version="1.0" encoding="UTF-8"?>
<bpmn:definitions xmlns:bpmn="http://www.omg.org/spec/BPMN/20100524/MODEL" xmlns:camunda="http://camunda.org/schema/1.0/bpmn" id="Definitions_1" targetNamespace="http://bpmn.io/schema/bpmn" exporter="Camunda Modeler" exporterVersion="1.7.2">
  <bpmn:process id="Process_1" isExecutable="true">
    <bpmn:task id="task" name="Task with error" />
    <bpmn:boundaryEvent id="errorBoundaryEvent" attachedToRef="task">
      <bpmn:errorEventDefinition errorRef="Error_0rrufgq" camunda:errorCodeVariable="statusCode" camunda:errorMessageVariable="message" />
    </bpmn:boundaryEvent>
  </bpmn:process>
  <bpmn:error id="Error_0rrufgq" name="Error_2adefbt" errorCode="404" />
</bpmn:definitions>`;

moddle.fromXML(withoutIdDef, (err, definition, context) => {
  console.log('ERROR IS ORPHAN', JSON.stringify(context.references, null, 2))
});

const withIdDef = `<?xml version="1.0" encoding="UTF-8"?>
<bpmn:definitions xmlns:bpmn="http://www.omg.org/spec/BPMN/20100524/MODEL" xmlns:camunda="http://camunda.org/schema/1.0/bpmn" id="Definitions_1" targetNamespace="http://bpmn.io/schema/bpmn" exporter="Camunda Modeler" exporterVersion="1.7.2">
  <bpmn:process id="Process_1" isExecutable="true">
    <bpmn:task id="task" name="Task with error" />
    <bpmn:boundaryEvent id="errorBoundaryEvent" attachedToRef="task">
      <bpmn:errorEventDefinition id="errorBoundaryEvent_definition" errorRef="Error_0rrufgq" camunda:errorCodeVariable="statusCode" camunda:errorMessageVariable="message" />
    </bpmn:boundaryEvent>
  </bpmn:process>
  <bpmn:error id="Error_0rrufgq" name="Error_2adefbt" errorCode="404" />
</bpmn:definitions>`;

moddle.fromXML(withIdDef, (err, definition, context) => {
  console.log('ERROR REFERRED', JSON.stringify(context.references, null, 2))
});
```

Not sure that I have foreseen all scenarios where this might cause a problem. But I have tried to add tests where applicable.


nikku:
What exactly are you trying to achieve and how does this PR address your issue?

paed01:
I'm building a bpmn execution engine that depends on `bpmn-moddle` to parse the BPMN2 xml. I use the Camunda Modeler to design the diagrams, simply because it is, imho, the best.

When the xml is transformed by `bpmn-moddle` you get the definitions and a list of references between elements.

Currently bound event definitions lacks id. The reference between the event definition and the defined error is present, but it is programmatically hard to find since there is no element id.

The PR addresses this by adding an id to event definitions when the tool selects the required type. But off course I might have misinterpreted how it works.

The error with name and code:
```
"Error_0rrufgq": {
  "$type": "bpmn:Error",
  "id": "Error_0rrufgq",
  "name": "Error_2adefbt",
  "errorCode": "404"
}
```

References when `eventDefinition` is **without** id:
```js
[
  {
    "element": {
      "$type": "bpmn:BoundaryEvent",
      "id": "errorBoundaryEvent",
      "eventDefinitions": [
        {
          "$type": "bpmn:ErrorEventDefinition"
        }
      ]
    },
    "property": "bpmn:attachedToRef",
    "id": "task"
  },
  { // The reference is present
    "element": {
      "$type": "bpmn:ErrorEventDefinition"
    },
    "property": "bpmn:errorRef",
    "id": "Error_0rrufgq"
  }
]
```

And if the `eventDefinition` has an id the reference list looks like this:
```js
[
  {
    "element": {
      "$type": "bpmn:BoundaryEvent",
      "id": "errorBoundaryEvent",
      "eventDefinitions": [
        {
          "$type": "bpmn:ErrorEventDefinition",
          "id": "errorBoundaryEvent_definition"
        }
      ]
    },
    "property": "bpmn:attachedToRef",
    "id": "task"
  },
  {
    "element": {
      "$type": "bpmn:ErrorEventDefinition",
      "id": "errorBoundaryEvent_definition" // ...it can be found
    },
    "property": "bpmn:errorRef",
    "id": "Error_0rrufgq"
  }
]
```

nikku:
So the underlying feature request is: Make sure an `EventDefinition` contains an ID?

paed01:
Yes, that is a more accurate description of what I'm trying to achieve. Thanks for making it clear. I struggled a bit with the commit message. Should I change the message?

nikku:
How exactly do you resolve event definitions referencing `Error` elements?

And why is looking into the reference list not enough? 

Provide us with a little bit more context; there might be a simple workaround suitable for you.

lib/features/modeling/ElementFactory.js
```diff
@@ -101,7 +101,9 @@ ElementFactory.prototype.createBpmnElement = function(elementType, attrs) {
 
   if (attrs.eventDefinitionType) {
     eventDefinitions = businessObject.get('eventDefinitions') || [];
-    newEventDefinition = this._moddle.create(attrs.eventDefinitionType);
+    newEventDefinition = this._moddle.create(attrs.eventDefinitionType, {
```
nikku: The right place to encode the _always needs ID constraint` would be [`BpmnFactory#L16`](https://github.com/bpmn-io/bpmn-js/blob/master/lib/features/modeling/BpmnFactory.js).

lib/features/modeling/BpmnFactory.js
```diff
@@ -94,4 +94,14 @@ BpmnFactory.prototype.createDiPlane = function(semantic) {
   });
 };
 
+BpmnFactory.prototype.createEventDefinition = function(type, parent) {
+  var element = this.create(type, {
```
nikku: This won't work, as an event may have more than one event definition; in that case, the ids would collide.

Instead, let `BpmnFactory` assign a random id.

nikku:
I understand this requirement:

> So the underlying feature request is: Make sure an `EventDefinition` contains an ID?

But I really don't understand _why_ you need that stuff / what you'd like to achieve with it:

* You get all boundary events via `context.elementsById` and could traverse them to find the error event definition
* You got the `errorEventDefinition` in your hand when inspecting the `context.references` property:

    ```javascript
    context.references.forEach(function(ref) {
      var element = ref.element;

      if (ref.property === 'bpmn:errorRef') {
        // element is the model element; inspect it,
        // traverse the BPMN tree from and so forth

        // this is the error I'm referencing
        console.log(element.errorRef);
      }
    });
    ```

In general `JSON` stringifying any model element will not serialize any hidden properties such as `$parent`, references and any other virtual pointers. Use `console.log(element)` or a JavaScript debugger to see what is actually inside a model element.

paed01:
Good point. I was somewhat aware of that but didn't check it out.

My engine only persists the `moddleContext` when saving state and not the entire XML so the hidden information will get lost. But that might not justify this PR. Up to the authors I guess.

nikku:
Will close this for now.

If you distribute your own BPMN modeler, you may provide a custom `BpmnFactory` with the changes you provided. That is only required, if you decide to depend on that feature.

I'd rather (personally) depend on the semantic structure of the document and parse event definitions and related errors from there. That's personal taste though :wink:.

paed01:
Your modeler i quite fine, thank you.

Working with JSON is more comfortable and you avoid the overhead of transforming the document. Especially since the engine is destined to execute on frontend.

PR #722, opened 2017-10-31
chore(renderer): Implemented oop paradigm for BpmnRenderer

felixlinker:
### Proposed Changes

* `lib/draw/BpmnRenderer.js` now has all functions implemented as prototype functions allowing for usefull inheritance

In my case, I'd like to be able to extend the `handlers`-map which wasn't possible before without making use of extensive copy'n'pasting.

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -46,439 +46,25 @@ function BpmnRenderer(eventBus, styles, pathMap, canvas, priority) {
 
   BaseRenderer.call(this, eventBus, priority);
 
-  var rendererId = RENDERER_IDS.next();
+  this.canvas = canvas;
+  this.pathMap = pathMap;
 
-  var textUtil = new TextUtil({
+  this.rendererId = RENDERER_IDS.next();
+
+  this.textUtil = new TextUtil({
     style: LABEL_STYLE,
     size: { width: 100 }
   });
 
-  var markers = {};
+  this.markers = {};
 
-  var computeStyle = styles.computeStyle;
+  this.computeStyle = styles.computeStyle;
 
-  function addMarker(id, options) {
-    var attrs = assign({
-      fill: 'black',
-      strokeWidth: 1,
-      strokeLinecap: 'round',
-      strokeDasharray: 'none'
-    }, options.attrs);
+  var self = this;
 
-    var ref = options.ref || { x: 0, y: 0 };
-
-    var scale = options.scale || 1;
-
-    // fix for safari / chrome / firefox bug not correctly
-    // resetting stroke dash array
-    if (attrs.strokeDasharray === 'none') {
-      attrs.strokeDasharray = [10000, 1];
-    }
-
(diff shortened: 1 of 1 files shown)


nikku:
I'd like to understand a little bit more about your case:

* Why would you like to extend the handler map?
* How do you plan to extend it?

felixlinker:
Sorry, I was a bit wrong in my op. I don't want to extend the `handlers`-map but the `PathMap`! I'd like to do something like this:

```js
MyRenderer.prototype.drawShape = function(parentGfx, element) {

  if (is(element, 'bpmn:Task') && shouldHaveCustomIcon(element)) {
    var task = this.handlers['bpmn:Task'](parentGfx, element);

    var pathData = this.pathMap.getScaledPath('CUSTOM_ICON', { /* ... */ });
    this.drawPath(parentGfx, pathData, { /* ... */ });

    return task;
  } else {
      return BpmnRenderer.prototype.drawShape.call(this, parentGfx, element);
  }
};
```

Without my commit, I'd be unable to call functions like `drawPath` as they're hidden in the constructor by closure.

nikku:
Thanks for clarifying. 

In general, as part of this library we distinguish between private (`_` prefixed) and public members of services. From what it looks like the stuff you're doing is more like _you better know what you're doing_. That said, all methods you expose should be private. Same goes for the member properties (i.e. _pathMap_ and friends). 

If it is just about changing the element decoration (the icon in this case), why not create an extension point that allows you to just hook into the decoration phase?

Example: 

```javascript
MyRenderer.prototype.decorateTask = function(parentGfx, element) {

  if (is(element, 'bpmn:Task') && shouldHaveCustomIcon(element)) {
    var pathData = pathMap.getScaledPath('CUSTOM_ICON', { /* ... */ });
    return this.drawPath(parentGfx, pathData, { /* ... */ });
  }
  
  return BpmnRenderer.prototype.decorateTask.call(this, parentGfx, element);
};
```

This is a narrowed down extension point that serves one well defined purpose. Plus, it would be easy to test that extension point, too.

felixlinker:
I currently don't see this working other than when accompanied by a new `decoratorHandlerMap`. One could than extend `BpmnRenderer.prototype.drawShape` like this:
```js

**Comment by nikku** [comment]
All right then.

My suggestion would be: 

* expose the minimal subset of functions you need with the renderer:

    ```javascript
    function Renderer(...) {

      ... // existing render code

      // internal API, exposed for extensions
      this._drawPath = drawPath;
      this._foo = foo;
      this._bar = bar;
    }
    ```

    Mind the usage of `_` to denote these are actually internal APIs that may break any time.

* if you want us to recognize that we break access to these internals, add test cases for them 😉 .

**Comment by felixlinker** [comment]
So that means you'd be ok with a PR that simply adds the following LOC?
```js
function BpmnRenderer(...) {
  /* ... */
  this._drawPath = drawPath;
}
```

nikku:
Yes. Plus think about, whether to add test cases or not.

PR #723, opened 2017-11-10
chore(renderer): exposed _drawPath in constructor

felixlinker:
### Proposed Changes

* The renderer now has the function `drawPath` exposed as `renderer._drawPath` which allows for implementing custom paths and rendering them.

I did decide to not add additional test cases as I don't have the expertise required to check for correctness by using the `tiny-svg` library.

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -342,6 +342,7 @@ function BpmnRenderer(eventBus, styles, pathMap, canvas, priority) {
 
     return path;
   }
+  this._drawPath = drawPath;
 
   function drawMarker(type, parentGfx, path, attrs) {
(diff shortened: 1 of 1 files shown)


nikku:
Closed via 0cb46183eb1c61be3d3d8acf6957d079a2990d91.

Checkout the [simple test case I've added](https://github.com/bpmn-io/bpmn-js/commit/0cb46183eb1c61be3d3d8acf6957d079a2990d91#diff-44c0dc66d77859ac94e34f4ca515b765R350), too.

felixlinker:
Alright! I was trying to implement a much more complicated test case by supplying an svg-object to the function and checking whether any path was appended correctly. But if this works - cool. Now I know :)

Issue #731, opened 2017-11-30
Contributing-Request / RendererUtil


felixlinker:
Hi there,

would you be okay with me moving all functions starting from [line 1811 in the Renderer](https://github.com/bpmn-io/bpmn-js/blob/master/lib/draw/BpmnRenderer.js#L1811) in a new module `lib/util/RendererUtil.js` analogously to `ModelUtil`, etc.? Then other people (in this case me 😉) could use them without copying them.

I'd just like to ask before making myself any work.

Kind regards
Felix

nikku:
Makes sense, go ahead.

A few hints:

* Add test cases for `RenderUtil` existing; otherwise we'll move / remove it any time without noticing

* Import the utilities from `RenderUtil` like this:

    ```javascript
    var foo = require('./RenderUtil').foo;
    ```

PR #732, opened 2017-12-01
chore(renderer): Created new module BpmnRenderUtil

felixlinker:
References #731 

### Proposed Changes

* Moved all functions starting from [line 1811 in the `Renderer`](https://github.com/bpmn-io/bpmn-js/blob/master/lib/draw/BpmnRenderer.js#L1811) to a new module `lib/util/RendererUtil.js`
* Test cases added checking for those functions to be defined

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -3,21 +3,32 @@
 var inherits = require('inherits'),
     isObject = require('lodash/lang/isObject'),
     assign = require('lodash/object/assign'),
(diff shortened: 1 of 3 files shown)


nikku:
Merged via https://github.com/bpmn-io/bpmn-js/commit/36a12396fd050243f1f9a5e0bf6a7c41f8a91002 with the following changes:

* moved the util to `draw/BpmnRenderUtil`, as no one should depend on it as _public api_

PR #741, opened 2017-12-13
feat(draw): render sequence flows behind tasks

### LINKED ISSUE #727: Render sequence flows behind other flow elements
As suggested in [this forum thread](https://forum.bpmn.io/t/links-overlapping-tasks/1904/4) we could render sequence flows behind other flow nodes, giving flow nodes slightly more priority and possibly improving the diagram look: 

sequenceflow-behind

__Tasks__

* [x] new ordering once moved / created
* [x] new ordering in drag groups
* [x] slight opacity on all flow elements
* [x] proper rendering order during import (is this possible _?_)

nikku:
This was surprisingly easy to implement.

* On import we simply use `parentIndex` to import sequence flows before all other elements
* On modeling the `orderingProvider` provides us all the tools we need to simply move sequence flows behind other flow nodes

In diagram-js, I had to ensure `parentIndex` is taken into account during GFX creation to make this feature work (cf. https://github.com/bpmn-io/diagram-js/commit/f5cf5b2aa930be3bdb26840548cb7aad1c5800ae). 

__Interaction Demo__

foo

Interaction note: As seen in the screencast a user is still able to interact with a selected sequence flow despite the fact that it is _behind_ the task.

Closes #727

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -50,6 +50,9 @@ var LABEL_STYLE = {
   fontSize: 12
 };
 
+var DEFAULT_FILL_OPACITY = .95,
+    HIGH_FILL_OPACITY = .35;
+
 
 function BpmnRenderer(eventBus, styles, pathMap, canvas, priority) {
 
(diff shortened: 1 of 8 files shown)


nikku:
Merged via 1e6186e3ce2e774daa2afa17ec378a49aacc948f.

PR #783, opened 2018-04-11
chore(BpmnSnapping): allow snapping for data associations

### LINKED ISSUE #782: No snapping to data object on initial connect
### Expected Behavior

When I create a connection with a data object as target, I would expect that it snaps to it.

### Actual Behavior

There is no snapping, except when reconnecting it.

data-objects

### Steps to reproduce the Behavior

1. Create a data object and a start event
2. Try to to connect the start event to the data object

pedesen:
Fixes #782 

### Proposed Changes

The changes in this PR allow snapping for DataInputAssociations and DataOutputAssociations, which makes snapping possible when connecting from and to DataStores and DataObjects.

### DIFF
--- a/lib/features/snapping/BpmnSnapping.js
+++ b/lib/features/snapping/BpmnSnapping.js
@@ -167,8 +167,13 @@ export default function BpmnSnapping(eventBus, canvas, bpmnRules, elementRegistr
       context.initialSourcePosition = context.sourcePosition;
     }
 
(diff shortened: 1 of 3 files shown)


nikku:
Merged via https://github.com/bpmn-io/bpmn-js/commit/fcd1e2f12c8b6f9ffe78322692295bcb5882f749.

Thanks for your contribution.

PR #789, opened 2018-05-03
fix(auto-place): fix infinite loop during auto-place

### LINKED ISSUE #788: Infinite loop in Auto-place after a BoundaryEvent in a corner
### Expected Behavior

Auto-placement of element after BoundaryEvent.

### Actual Behavior

Infinite loop caused by a NaN.

### Steps to reproduce the Behavior

Infinite loop

Error seems to be in ./lib/features/auto-place/AutoPlaceUtil.js in L54, when orientation is set to 'top-right' or another corner, verticalDistances[orientation] returns NaN.
It then create an infinite loop.

nboissel:
Fix NaN return while autoplace an element after a BoundaryEvent in a corner.

Fixes #788

### Proposed Changes
Replace dict by function that computes vertical distance of an element after a boundary placed in a corner.

### DIFF
--- a/lib/features/auto-place/AutoPlaceUtil.js
+++ b/lib/features/auto-place/AutoPlaceUtil.js
@@ -42,16 +42,19 @@ export function getFlowNodePosition(source, element) {
     }
   }
 
-  var verticalDistances = {
-    left: 0,
-    right: 0,
-    top: -1 * rowSize,
-    bottom: rowSize
-  };
+  function getVerticalDistance(orient) {
+    if (orient.indexOf('top') != -1) {
+      return -1 * rowSize;
+    } else if (orient.indexOf('bottom') != -1) {
+      return rowSize;
+    } else {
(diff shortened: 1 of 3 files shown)


nikku:
Do you mind adding a test case that verifies the loop is fixed, too.

Should be as straight forward, as we already got good auto-place test coverage in [`AutoPlaceSpec`](https://github.com/bpmn-io/bpmn-js/blob/master/test/spec/features/auto-place/AutoPlaceSpec.js#L199).

nboissel:
Sure.
I add this boundary on an existing task in the test process. Hope this will be fine (this task was not used in the test yet)

nikku:
Merged your changes via https://github.com/bpmn-io/bpmn-js/commit/9379abafcc0b3c75e70bb75fe64d4d670c72a0c0. Thanks for both, the bug report and your fix.

PR #792, opened 2018-05-09
chore(contextPadProvider): adjust popupMenu method calls

pedesen:
Since breaking changes were introduced in diagram-js
popupMenu methos calls need to be adjusted

* call popupMenu#open with element, id and position as params
* call popupMenu#isEmpty with element and providerId as params
* remove popupMenu#create call

Related to bpmn-io/diagram-js#253


nikku:
Merged via https://github.com/bpmn-io/bpmn-js/commit/17896df65cdd940e669494e8411528fb213337f2.

PR #800, opened 2018-05-22
feat(replace): Keep timer boundary config when just switching interrupt

### LINKED ISSUE #799: Boundary TimerEventDefinitions config lost when switching between interrupting and non-interrupting
Hello,

The title is quite self explanatory. Configuration of boundary timer is lost when switching between interrupting and non-interrupting as its configuration is embedded in a child element and not in attributes.

Boundary timers can keep their configuration since it's just a small change in behavior.

My PR is coming for this if you also consider this as an issue.

nboissel:
TimerEventDefinitions loose its configuration (which is a child element) when switching between interrupting and not interrupting.

Fixes #799

### Proposed Changes

* Copy timerEventDefinition child if switch appends between two timerBoundary

### DIFF
--- a/lib/features/replace/BpmnReplace.js
+++ b/lib/features/replace/BpmnReplace.js
@@ -144,6 +144,30 @@ export default function BpmnReplace(
     // initialize custom BPMN extensions
     if (target.eventDefinitionType) {
       newElement.eventDefinitionType = target.eventDefinitionType;
+
+      // Keep timer event def config if switching between interrupting and non-interrupting
+      if (oldBusinessObject.eventDefinitions) {
+        var oldEventDef = oldBusinessObject.eventDefinitions[0];
+        if (oldEventDef &&
+            is(oldEventDef, 'bpmn:TimerEventDefinition') &&
+            target.eventDefinitionType === oldEventDef.$type) {
+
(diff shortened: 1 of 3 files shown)


nboissel:
After further investigation, I must miss something because the timer configuration is well kept but there is a second empty <timerEventDefinition /> in the XML when I save it.
Any idea about that? Do you know where it could be created?

nikku:
Looking at your changes and the overall matter right now. Trying to figure out, what's up :sweat_smile:.

nikku:
As part of _replace_ we clone properties (cf. [`BpmnReplace#L147`](https://github.com/bpmn-io/bpmn-js/blob/c8db48a8272cb3bf762ff6d5a5b94ff1f1082d05/lib/features/replace/BpmnReplace.js#L147)) but excluded the event definitions until now.

The proper fix ensures we clone the definitions, if appropriate (cf. https://github.com/bpmn-io/bpmn-js/commit/c8db48a8272cb3bf762ff6d5a5b94ff1f1082d05).

nikku:
Thanks for the PR.

The appropriate fix will be merged via https://github.com/bpmn-io/bpmn-js/pull/818.

PR #801, opened 2018-05-24
Text rendering unification

nikku:
This factors out text rendering and dimension computation into a `TextRenderer` service.

As a result, we get rid of a little bit of code duplication.


nikku:
Restored commit: https://github.com/bpmn-io/bpmn-js/commit/4bb270f19279db40f9cc3c179e09ee3a9a114e7c

PR #842, opened 2018-08-01
fix(chore): fix main file location, ensure compatibility with nodejs

huochunpeng:
The current setup uses implicit default main file `index.js` which uses es6 native module `import` syntax, and directly loads from source file (`lib` folder) which uses native module `import` syntax too.

The whole setup bypasses `dist` file prepared by rollup.

The reason that this setup still works in webpack, is webpack supports es6 native module `import` syntax. We don't have the luck if we use the main file in nodejs or browser environments directly (both currently do not support `import` syntax by default).

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,0 @@
-export {
-  default
-} from './lib/Viewer';
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "bpmn-js",
   "version": "2.4.1",
   "description": "A bpmn 2.0 toolkit and web modeler",
+  "main": "dist/bpmn-viewer.production.min.js",
   "scripts": {
     "all": "run-s lint test distro test:distro",
     "lint": "eslint .",
--- a/test/spec/ViewerSpec.js
+++ b/test/spec/ViewerSpec.js
@@ -2,8 +2,6 @@ import TestContainer from 'mocha-test-container-support';
 
 import Diagram from 'diagram-js/lib/Diagram';
 
-import ViewerDefaultExport from '../../';
-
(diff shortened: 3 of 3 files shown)


nikku:
What is your setup / which setups would you like to support?

huochunpeng:
The linked issue above uses aureli-cli which uses a requirejs based build, it supports normal commonjs/umd/amd code which is shipped by most npm packages.

The current setup stops nodejs with a simple `var BpmnJS = require('bpmn-js');`.
But `var BpmnJS = require('bpmn-js/dist/bpmn-viewer.production.min');` works fine after properly preparing `jsdom`.

nikku:
I guess aurelia is using browserify to consume libraries and bundle things for the browser. 

Did you consider adding the respective `babelify` transform to translate ES module import / export declarations to cjs?

nikku:
The alternative is to use the workaround you provide.

huochunpeng:
I understand, this is not a big issue for using bpmn in aurelia. After all, it's your call whether to make it smoother to nodejs env and other bundlering tool other than webpack.

nikku:
> After all, it's your call whether to make it smoother to nodejs env and other bundlering tool other than webpack.

We came from CommonJS and switched to ES modules for reasons such as smaller bundle sizes / tree shaking. My guess is that eventually, all decent bundlers _and_ NodeJS will understand ES modules, as it is the new standard for JavaScript modules.

Before we reach this point, people that are forced into AMD / cjs environments with no ability to control the actual transpilation should resort to the workaround you mentioned:

> `var BpmnJS = require('bpmn-js/dist/bpmn-viewer.production.min')` works fine after properly preparing jsdom.

PR #873, opened 2018-10-16
784 no boundary event after event based gateway

pinussilvestrus:
Implemented a Bpmn rule that it shouldn't be possible to attach a BoundaryEvent on a ReceiveTask following a  EventBasedGateway in two different ways:

- It isn't possible to attach a BoundaryEvent on a ReceiveTask which is already connected after a EventBasedGateway (BpmnRule)
- Deletion of attached BoundaryEvent after connecting a EventBasedGateway to a ReceiveTask (Behavior)

Closes https://github.com/camunda/camunda-modeler/issues/784

### DIFF
--- a/lib/features/modeling/behavior/ConnectEventBasedGatewayBehavior.js
+++ b/lib/features/modeling/behavior/ConnectEventBasedGatewayBehavior.js
@@ -1,0 +1,47 @@
+import inherits from 'inherits';
+
+import CommandInterceptor from 'diagram-js/lib/command/CommandInterceptor';
+
+import { is } from '../../../util/ModelUtil';
+
+import {
+  filter
+} from 'min-dash';
+
+
+/**
+ * Behavior for deleting boundaries from receive task after connecting them with event based gateway
+ */
+export default function ConnectEventBasedGatewayBehavior(eventBus, modeling) {
+
+  CommandInterceptor.call(this, eventBus);
+
+  function extractBoundaryEvents(element) {
+    return filter(element.attachers, function(attacher) {
+      return is(attacher, 'bpmn:BoundaryEvent');
+    });
+  }
+
+  this.postExecute('connection.create', function(context) {
+    var source = context.context.source,
+        target = context.context.target,
+        boundaries = extractBoundaryEvents(target);
+
+    if (
+      is(source, 'bpmn:EventBasedGateway') &&
+      is(target, 'bpmn:ReceiveTask') &&
+      boundaries.length > 0
+    ) {
+      modeling.removeElements(boundaries);
+    }
+
+  });
+}
+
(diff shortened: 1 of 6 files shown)


pinussilvestrus:
after review session with @barmac another case should be considered:
- change type of gateway to eventbasedgateway which is still connected to receive task with boundary --> boundary should be removed

lib/features/modeling/behavior/ConnectEventBasedGatewayBehavior.js
```diff
@@ -0,0 +1,51 @@
...
+import {
+  filter
+} from 'min-dash';
+
+
+/**
+ * Behavior for deleting boundaries from receive task after connecting them with event based gateway
+ */
+export default function ConnectEventBasedGatewayBehavior(
+    eventBus, modeling, bpmnRules) {
```
barmac: `bpmnRules` is actually never used, so I think we should remove it before merge.

lib/features/modeling/behavior/ConnectEventBasedGatewayBehavior.js
```diff
@@ -0,0 +1,51 @@
...
+    );
+  }
+
+  this.postExecute('connection.create', function(context) {
+    var source = context.context.source,
+        target = context.context.target;
+
+    if (is(source, 'bpmn:EventBasedGateway') &&
+        isReceiveTaskWithBoundary(target)) {
+      modeling.removeElements(extractBoundaryEvents(target));
```
barmac: So we invoke `extractBoundaryEvents` twice: first to check if the element is a `ReceiveTask` with `BoundaryEvent`s, and then to filter out all the `BoundaryEvent`s and remove them. Why don't we filter the events once and then remove them? :)

test/spec/features/modeling/behavior/ConnectEventBasedGatewayBehaviorSpec.js
```diff
@@ -0,0 +1,36 @@
...
+import coreModule from 'lib/core';
+
+
+describe('features/modeling/behavior - connect event based gateway to receive task with boundary', function() {
+
+  var testModules = [ coreModule, modelingModule ];
+
+  var diagramXML = require('./ConnectEventBasedGatewayBehavior.bpmn');
+
+  beforeEach(bootstrapModeler(diagramXML, { modules: testModules.concat(modelingModule) }));
```
barmac: Why do we add `modelingModule` second time here? The tests seem to work without it.

lib/features/modeling/behavior/ConnectEventBasedGatewayBehavior.js
pinussilvestrus: adjust and rebased it :)

PR #883, opened 2018-10-25
Update dev

nikku:
This fixes our CI build and performs some general clean up.

To run code reliably on CI I've pinned the node_js version to use to 10: https://github.com/bpmn-io/bpmn-js/commit/19904f9c5d4f06a16c108dd1e19a8252964e16b6.

Bumping the dev dependencies, including rollup and friends gives us a minor improvement in the production bundle size, too:

#### Before

```sh
> du -hb dist/*
1189216 dist/bpmn-modeler.development.js
429340  dist/bpmn-modeler.production.min.js
419515  dist/bpmn-navigated-viewer.development.js
164787  dist/bpmn-navigated-viewer.production.min.js
422175  dist/bpmn-viewer.development.js
161804  dist/bpmn-viewer.production.min.js
```

#### After

```sh
> du -hb dist/*
1204854 dist/bpmn-modeler.development.js
426345  dist/bpmn-modeler.production.min.js
435241  dist/bpmn-navigated-viewer.development.js
163617  dist/bpmn-navigated-viewer.production.min.js
441191  dist/bpmn-viewer.development.js
160686  dist/bpmn-viewer.production.min.js
```

PR #884, opened 2018-10-25
chore(project): detect unresolved lint rules

nikku:
This ensures we don't accidently require files in the lib directory
via the global `lib` import. That stuff works during tests but does
not work in the final bundle.

Considerations: This plugin adds some runtime overhead to our linting
process.

Related to https://github.com/bpmn-io/diagram-js/pull/292.

### DIFF
--- a/lib/.eslintrc
+++ b/lib/.eslintrc
@@ -1,0 +1,8 @@
+{
+  "plugins": [
+    "import"
+  ],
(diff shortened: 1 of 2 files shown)


nikku:
Note to reviewer: This fails due to https://github.com/bpmn-io/bpmn-js/pull/883 not being merged.

PR #885, opened 2018-10-26
Adjust keybindings to new api

barmac:
This PR adjusts `BpmnKeyBindings` to a breaking change introduced in `diagram-js` with merge of [#289](https://github.com/bpmn-io/diagram-js/pull/289) as well as replaces deprecated `KeyboardEvent.keyCode` with `KeyboardEvent.key`.

### DIFF
--- a/test/util/KeyEvents.js
+++ b/test/util/KeyEvents.js
@@ -1,8 +1,9 @@
-export function createKeyEvent(element, code, ctrlKey) {
+export function createKeyEvent(element, key, ctrlKey) {
   var e = document.createEvent('Events') || new document.defaultView.CustomEvent('keyEvent');
 
-  e.keyCode = code;
-  e.which = code;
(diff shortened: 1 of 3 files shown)


test/util/KeyEvents.js
```diff
@@ -1,8 +1,9 @@
-export function createKeyEvent(element, code, ctrlKey) {
+export function createKeyEvent(element, key, ctrlKey) {
   var e = document.createEvent('Events') || new document.defaultView.CustomEvent('keyEvent');
 
-  e.keyCode = code;
-  e.which = code;
+  e.key = key;
```
nikku: If we decide to only look into `event.key` we should probably not mock the other properties, right?

barmac:
Closed via [#887](https://github.com/bpmn-io/bpmn-js/pull/887)

PR #886, opened 2018-10-26
376 add keyboard shortcuts to move elements

barmac:
This PR activates `KeyboardNavigation` and `MoveSelection` features added in `diagram-js` per default.

Note to reviewer: 
Includes changes in https://github.com/bpmn-io/bpmn-js/pull/885. Review the linked PR first.

Closes #376

PR #887, opened 2018-10-26
Editor action refactoring

nikku:
This builds upon https://github.com/bpmn-io/diagram-js/pull/295.

Refactors the editor actions to be independent from actual editor (or viewer) features, as discussed with @barmac.

Core features:

* only bind to actions in EditorActions which are implemented via features
* only register keyboard shortcuts for actions that are implemented via features
* explicitly include modules in `NavigatedViewer` and `Modeler`
* use key event exposed via keyEvent property

This superseeds https://github.com/bpmn-io/bpmn-js/pull/886, if merged.


barmac:
Added `diagram-js@3.0.0` dependency

PR #928, opened 2019-01-03
Update CONTRIBUTING.md

szurilo:

Closes #

Project structure is not okay without this change.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -39,6 +39,7 @@ cd bpmn.io
 
 git clone git@github.com:bpmn-io/diagram-js.git
 (cd diagram-js && npm i)
+cd ..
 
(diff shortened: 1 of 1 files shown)


nikku:
You're on Windows, right? That should explain why the setup instructions don't work for you.

In a *nix like environment `(cd foo && bar)` has the semantics 

* move to `foo`
* execute `bar`
* go back

szurilo:
Are you sure? I just tried it on CentOS and with these commands the CWD is /foo
cd /
mkdir foo
cd foo && touch bar

nikku:
You must execute the statement as shown above, on a single line: `(cd foo && bar)`.

nikku:
The brackets matter.

szurilo:
Ahh, you are right, thanks for the clarification. I assumed the brackets are there for commenting only :)

PR #931, opened 2019-01-04
feat: Autoresizing labels based on its content size

szurilo:

Closes #

This is actually coming from this forum post: https://forum.bpmn.io/t/name-of-textannotation-resize-eventhandler/2646/4 because Niklas_Kiefer suggested to create a PR.
The implementation is not mine as you can see in the forum so all credit goes to jclaxton. I am just doing this pull request bec. I also need this feature.

However I ran npm test and see that 4 tests are failing after these modifications.
These:
should expand width
should reduce height
should remain the same
should keep label position

If you ask me I would delete those test cases from now on, but I am open to any other proposal how to fix the tests.

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -471,8 +471,8 @@ export default function BpmnRenderer(
   function renderExternalLabel(parentGfx, element) {
     var semantic = getSemantic(element);
     var box = {
-      width: 90,
-      height: 30,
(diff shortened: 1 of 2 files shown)


nikku:
Could you provide a test case that demonstrates the scenario you'd like to support?

barmac:
Closed because of no activity for over 2 weeks. This can be always reopened, though.

PR #948, opened 2019-03-18
improvement: generate native javascript module distributions

newportd:
Hello, this pull request will generate native JavaScript module distributions that can be directly included using e.g. import BpmnJS from '/bpmn-modeler.esm.development.js';

My use case is to be able to use the distribution directly without requiring a separate bundler.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7481,6 +7481,79 @@
         "rollup-pluginutils": "^2.0.1"
       }
     },
+    "rollup-plugin-terser": {
+      "version": "4.0.4",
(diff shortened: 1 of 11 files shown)


nikku:
Thanks for your contribution.

Could you elaborate why we want that bundle? What is the benefit is over simply including the pre-built distribution?

```javascript
import BpmnModeler from 'bpmn-js/dist/bpmn-modeler.production.min.js';
```

nikku:
Closing this issue due to [missing feedback](https://github.com/bpmn-io/bpmn-js/pull/948#issuecomment-476112339). Happy to re-open it, if you provide us with the desired input.

PR #966, opened 2019-04-03
Task resizing

ElCondor1969:
This PR wants to add the ability to resize tasks in the diagram.
The functionality is enabled through a special option:

`
var viewer = new BpmnJS({
  container: 'body', 
  taskOptions: {
    resizeEnabled: true
  }
});
`

This feature is essential if you consider that through the editor BPMN-JS you can also open diagrams written with other editors.
You may need to adjust the size of tasks altered by other editors.

### DIFF
--- a/lib/features/rules/BpmnRules.js
+++ b/lib/features/rules/BpmnRules.js
@@ -39,15 +39,17 @@ import {
 /**
  * BPMN specific modeling rule
  */
-export default function BpmnRules(eventBus) {
+export default function BpmnRules(eventBus, taskOptions) {
   RuleProvider.call(this, eventBus);
+  this.taskOptions=taskOptions || {};
 }
 
 inherits(BpmnRules, RuleProvider);
 
-BpmnRules.$inject = [ 'eventBus' ];
+BpmnRules.$inject = [ 'eventBus', 'config.taskOptions' ];
 
 BpmnRules.prototype.init = function() {
+  var me=this;
 
   this.addRule('connection.start', function(context) {
     var source = context.source;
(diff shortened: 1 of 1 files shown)


pinussilvestrus:
Hi @ElCondor1969,
first thanks for your contribution. Right now, it isn't that clear to me, what goal this pull request is going to achieve. Can you explain an _general_ scenario on why we want to have this kind of option? Tasks are already resizable inside the `Modeler`. Your implementation seems to deal with very specific cases, thats why I don't see to add this as an official option.

Extra Note: Please check our [Contribution guidelines](https://github.com/bpmn-io/bpmn-js/blob/master/.github/CONTRIBUTING.md).

ElCondor1969:
Hi, thank you for your reply.
I'll try to be clearer.
Let's draw this simple diagram:

image

Without enabling the option introduced by this PR, by selecting the task, we see that there are no "grips" at the four corners of the task that allows us to resize, by the mouse, the rectangle of the task.

By enabling the option of this PR, once the task is selected, the grips appear at the four corners of the task:

image

This gives us the possibility, with the mouse, to resize the rectangle of the task:

image

I hope now to have explained well the purpose of this PR.
Bye.

pinussilvestrus:
Ah I get it!

Resizing tasks is a often asked feature of our modeling toolkits. We decided to not offer the resize function for that kind of shapes (cf. this[ forum thread for rationale](https://forum.bpmn.io/t/resizable-tasks/480/2?u=nikku)). So I don't think this Pull request will be merged.

One thing you could do is to move this feature to a custom extension for `bpmn-js` which allows it and others can include it in their Modelers (like happen [inside this extension](https://github.com/bpmn-io/bpmn-js-nyan))

ElCondor1969:
I got it.
I'll write a custom extension like you suggest me.
Thank you.

PR #972, opened 2019-04-10
fix(draw): render hex and rgb encoded color markers

### LINKED ISSUE #981: Custom fill and stroke colors make markers disappear
__Describe the Bug__

When you use more complex color values for defaultFillColor and defaultStrokeColor (e.g. rgba() function) the renderer will use theses values in the ids for markers. As the ids then contain characters not valid for urls and element references in svg the marker cannot be linked. This finally results in missing markers in the svg image (e.g. missing arrow heads for sequence flows).

Afaics this is the only place where complex color values do not work.

__Steps to Reproduce__

Define defaultFillColor and defaultStrokeColor like this:
```
bpmnRenderer: {
    defaultFillColor: "rgba(255,255,255,0.95)",
    defaultStrokeColor: "rgba(30,30,30,1)"
}
```

https://jsfiddle.net/y1nq86sm/

__Expected Behavior__

* [ ] Escape colors used in marker ids so that resulting IDs are valid

__Environment__

 - Browser: Chrome
 - OS: Win 10
 - Library version: 3.3.1

nikku:
Until now we did not properly sanitize the IDs we create for colored
markers.

This would make the marker retrival fail with fill and stroke colors
encoded in hex and rgb(a, b, c) form.

This improves the situation by removing forbidden chars from the marker
ID.

Related to camunda/camunda-modeler#1342
Closes #981


nikku:
Not ready for review yet. I discovered that we still got the same issue for message flow and association ends.

nikku:
Ready for review.

PR #975, opened 2019-04-12
Adjust height of pool when toggling collapse/expand button

gustavjf:

Closes camunda/camunda-modeler#1210


barmac:
Change of the width makes us run into new issue:

connection

It seems like we don't layout the connections after replacement.

nikku:
Can we simply keep the width _as is_?

gustavjf:
Here's the result when keeping the width as-is:

bpmn-js#975

PR #977, opened 2019-04-16
fix(BpmnTreeWalker): pass context to unhandled process handler

barmac:
Previously, we wouldn't pass context to `contextual` which didn't make any sense. This was created as part of https://github.com/bpmn-io/bpmn-js/pull/967 or `bpmn-js/select-diagram`.


nikku:
Could you add a test case that reproduces the now fixed behavior?

nikku:
And can we make it `fix(import/BpmnTreeWalker)`?

barmac:
Sure.

barmac:
Done.

PR #980, opened 2019-04-16
Removed fill and stroke from id for markers

FlexiGit:
The fill and stroke properties are derived from defaultFillColor and defaultStrokeColor. If you use more complex colors for these values (e.g. rgba(x, x, x, x) which is perfectly ok for the color values in the resulting svg), the resulting id contains characters which are not valid in urls (space and comma in this example). And following that, they cannot be used as reference to the markers definition in the svg.

Closes #

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -124,7 +124,7 @@ export default function BpmnRenderer(
   }
 
   function marker(type, fill, stroke) {
-    var id = type + '-' + fill + '-' + stroke + '-' + rendererId;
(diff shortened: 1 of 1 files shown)


FlexiGit:
The suggested fix would produce new errors.

nikku:
Have a look at https://github.com/bpmn-io/bpmn-js/pull/972/commits/e7b1e535f08e252d5ce1d4c553d7cc519b115584 for the actual fix. It still needs additional work for other makers (message flow, association, ...) so not merged yet.

PR #988, opened 2019-04-18
718 properly layout on reconnect

### LINKED ISSUE #718: Properly layout on connection end re-connection
This should avoid situations like this after the source or target of a shape got re-connected: 

image

__Tasks__

* [ ] Layout connection while reconnecting
* [ ] Layout connection while connecting

barmac:
Test adjustments according to https://github.com/bpmn-io/diagram-js/pull/320

Requires https://github.com/bpmn-io/diagram-js/pull/320

Closes #718


nikku:
Works like a charm. We could spend a little bit of time to look into these label layouting related issues, too.

I was able to discover a quirk myself while playing around / testing the feature.

PR #1004, opened 2019-04-29
feat(modeling): adjust default collapsed participant height to 60px

nikku:
This is a follow up to https://github.com/bpmn-io/bpmn-js/pull/975.

It sets the default height of collapsed pools to `60px` to better align with the behavior of other tools.


nikku:
When talking about _other tools_ I looked into Signavio, specifically.

nikku:
Done, cf. https://github.com/bpmn-io/bpmn-js/pull/1004/commits/abe3fc6e68ae106af02f4d3158bb637b4ad8813f.

PR #1006, opened 2019-04-29
Disallow multiple incoming sequence flows to target events of event-based gateway

gustavjf:

Relates to camunda/camunda-modeler#637

This updates an existing BPMN rule and adds new modeling behavior for event-based gateways.

When disallowing multiple incoming sequence flows to target events of an event-based gateway, there are three scenarios that are covered here:

Scenario 1: Disallowing a connection to an event-based target when the source is not an event-based gateway.

camunda-modeler#637_01

Scenario 2: Removing existing sequence flows of event-based targets when connecting from an event-based gateway.

camunda-modeler#637_02

Scenario 3: When a user replaces a gateway, that is already connected to event-based targets, with an event-based gateway, then the existing incoming sequence flows of the targets, which do not belong to the newly replaced event-based gateway, are removed.

camunda-modeler#637_05

A fourth scenario which is not covered here would be preventing the user from connecting an event-based gateway to a target which already has an incoming connection from that same event-based gateway. This will be looked at as a separate case.

Instead, the existing connection would be implicitly removed:

camunda-modeler#637_03


lib/features/modeling/behavior/EventBasedGatewayBehavior.js
```diff
@@ -0,0 +1,34 @@
...
+  // remove existing connection before connecting from event-based gateway
+  this.preExecuted('connection.create', function(event) {
+
+    var source = event.context.source,
+        target = event.context.target,
+        oldConnections = target.incoming;
+
+    if (
+      is(source, 'bpmn:EventBasedGateway') &&
+      target.incoming.length > 0
```
gustavjf: This handles multiple connections, despite the [test](https://github.com/bpmn-io/bpmn-js/pull/1006/commits/eb4aa6f354888598c4fdae72a68c1b4ef99c8b31#diff-c08486b559e39e159cf3632d833d3fc9R19) checking only for single connections.

This is because after ddc5c02, multiple connections are disallowed. However, it may be that a diagram is imported which was created with a tool that does not disallow multiple connections, so better to cover the case with multiple connections.

barmac:
Screenshot 2019-05-02 at 15 14 53

Screenshot 2019-05-02 at 15 13 02

It seems like somehow we don't remove the connection properly when we replace the gateways with event based gateways.

gustavjf:
@barmac Thanks, I've added tests and new behavior for dealing with the replacement scenario you mentioned. PR description has been updated to demonstrate this.

Could you have another look? Hopefully we've got it this time 👍

test/spec/features/modeling/behavior/EventBasedGatewayBehaviorSpec.js
```diff
@@ -0,0 +1,97 @@
...
+    ));
+
+  });
+
+
+  describe('when replacing with event based gateway', function() {
+
+    beforeEach(bootstrapModeler(
+      diagramXML,
+      { modules: [ coreModule, modelingModule, replaceModule ] }
```
barmac:
```suggestion
      diagramXML, { 
        modules: [ 
          coreModule,
          modelingModule,
          replaceModule 
        ] 
      }));
```

barmac: Same in other places for consistence's sake.

lib/features/rules/BpmnRules.js
```diff
@@ -896,3 +905,29 @@ function canCopy(collection, element) {
...
+  var result = false;
+
+  for (var i = 0; i < connections.length; i++) {
+    if (isOutgoingEventBasedGatewayConnection(connections[i])) {
+      result = true;
+      break;
+    }
+  }
+
+  return result;
```
barmac: 923-932 could be replaced simply with `connections.some(isOutgoingEventBasedGatewayConnection)`. It's implemented in IE11.

lib/features/modeling/behavior/EventBasedGatewayBehavior.js
```diff
@@ -0,0 +1,73 @@
...
+  this.preExecuted('shape.replace', function(event) {
+
+    var newShape = event.context.newShape,
+        newShapeOutgoingConnections = newShape.outgoing;
+
+    if (!is(newShape, 'bpmn:EventBasedGateway')) {
+      return;
+    }
+
+    newShapeOutgoingConnections
```
barmac: This part should be refactored as it's really difficult to understand. For sure we don't need `map` here. Consider assigning step results to variables.

lib/features/modeling/behavior/EventBasedGatewayBehavior.js
```diff
@@ -0,0 +1,73 @@
...
+        existingIncomingConnections = target.incoming.slice();
+
+    if (
+      is(source, 'bpmn:EventBasedGateway') &&
+      target.incoming.length > 0
+    ) {
+
+      existingIncomingConnections
+        .filter(function(connection) {
+          return is(connection, 'bpmn:SequenceFlow');
```
barmac: We check this condition multiple times in the file so why don't we have a helper for this? It will increase the readibility of the code :)

lib/features/modeling/behavior/EventBasedGatewayBehavior.js
```diff
@@ -0,0 +1,73 @@
...
+   */
+  this.preExecuted('connection.create', function(event) {
+
+    var source = event.context.source,
+        target = event.context.target,
+        existingIncomingConnections = target.incoming.slice();
+
+    if (
+      is(source, 'bpmn:EventBasedGateway') &&
+      target.incoming.length > 0
```
barmac:
```suggestion
target.incoming.length
```

lib/features/modeling/behavior/EventBasedGatewayBehavior.js
barmac:
```suggestion
filter(isSequenceFlow)
```

PR #1027, opened 2019-05-13
feat(translations): add new keywords

daniloakamine:
When I use translations there are some tooltips on my pallet that are still in English like `StartEvent`, `EndEvent` or `Task`.
It's a little bit weird because the phrase begins in my language and ends in other language.
The reason for that is because there is a generic keyword `Create {type}`.
I'm suggesting create specific keywords for these cases.

### DIFF
--- a/lib/features/context-pad/ContextPadProvider.js
+++ b/lib/features/context-pad/ContextPadProvider.js
@@ -313,7 +313,8 @@ ContextPadProvider.prototype.getContextPadEntries = function(element) {
       assign(actions, {
         'append.end-event': appendAction(
           'bpmn:EndEvent',
-          'bpmn-icon-end-event-none'
+          'bpmn-icon-end-event-none',
+          translate('Append EndEvent')
         ),
         'append.gateway': appendAction(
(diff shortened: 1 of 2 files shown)


nikku:
Thanks for your contribution.

You can already [customize the way translations work](https://github.com/bpmn-io/bpmn-js-examples/tree/master/i18n) to suite your needs. Did you consider that?

Because of that I'm inclined to not add additional translatable keywords.

daniloakamine:
Thanks for answering.
Actually, we have already customized in order to translate all of the `{types}`.
However, I would rather translate the full phrases instead of words.

Another point is that there are already some translations to `Create Gateway` or `Create Pool/Participant`. Why are other types different and use a generic way `Create {type}`?

It seems a little weird for me.
For `Create Gateway` we should use the default translation, but for `Create StartEvent` we have to customize?

Thanks again.

nikku:
This is going to be a breaking change for all existing translations.

That is another reason why I was hesitant to accept this PR.

Should be fine though, as we're going to cut a new major release next anyway.

PR #1039, opened 2019-05-21
Add subprocess with start event included

gustavjf:

Related to camunda/camunda-modeler#1242
Dependent on bpmn-io/diagram-js#347

This automatically adds a start event to an expanded subprocess when:

1. Creating an expanded subprocess from the palette.

camunda-modeler#1242_02

2. Replacing a task with an expanded subprocess.

camunda-modeler#1242_01

### DIFF
--- a/lib/features/modeling/behavior/SubProcessStartEventBehavior.js
+++ b/lib/features/modeling/behavior/SubProcessStartEventBehavior.js
@@ -1,0 +1,69 @@
+import inherits from 'inherits';
+
+import CommandInterceptor from 'diagram-js/lib/command/CommandInterceptor';
+
+import { is } from '../../../util/ModelUtil';
+import { isExpanded } from '../../../util/DiUtil.js';
+
+/**
+ * Add start event child by default when creating an expanded subprocess
+ * with create.start or replacing a task with an expanded subprocess.
+ */
+export default function SubProcessStartEventBehavior(eventBus, modeling) {
+  CommandInterceptor.call(this, eventBus);
+
+  eventBus.on('create.start', function(event) {
+    var shape = event.context.shape,
+        hints = event.context.hints;
+
+    hints.shouldAddStartEvent = is(shape, 'bpmn:SubProcess') && isExpanded(shape);
+  });
+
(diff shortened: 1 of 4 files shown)


nikku:
Looks good.

lib/features/modeling/behavior/SubProcessStartEventBehavior.js
```diff
@@ -0,0 +1,80 @@
...
+ * with create.start or replacing a task with an expanded subprocess.
+ */
+export default function SubProcessStartEventBehavior(eventBus, modeling) {
+  CommandInterceptor.call(this, eventBus);
+
+  eventBus.on('create.start', function(event) {
+    var shape = event.context.shape,
+        hints = event.context.hints;
+
+    if (
```
nikku: Shouldn't this be something like hints.shouldAddStartEvent = is('subporcess') && isExpanded(shape)?

That piece of code will set shouldaddStartEvent for every task.

lib/features/modeling/behavior/SubProcessStartEventBehavior.js
```diff
@@ -0,0 +1,80 @@
...
+
+    hints.shouldAddStartEvent = true;
+  });
+
+  this.postExecuted('shape.create', function(event) {
+    var shape = event.context.shape,
+        hints = event.context.hints,
+        position;
+
+    if (
```
nikku: Why do we need to double check here rather than simply checking `hints.shouldAddStartEvent`?

lib/features/modeling/behavior/SubProcessStartEventBehavior.js
gustavjf: Good point, that simplifies things.

lib/features/modeling/behavior/SubProcessStartEventBehavior.js
gustavjf: You're right, just checking `hints.shouldAddStartEvent` is sufficient here.

I was still thinking of the other cases where `shape.create` is invoked but that won't be an issue with the new hints flag since it will be undefined in those situations.

PR #1040, opened 2019-05-22
Group direct editing

### LINKED ISSUE #955: Enable directEditing for Groups

Right now it is not possible to edit a group's name via directEditing.

Make it possible to edit a group's name directly in the diagram.

__Additional context__

* A group's name is defined by a referenced `categoryValue`
* If not available, the Category should be created automatically

---

Child of #343

pinussilvestrus:
Closes #955 

* Moves Category + Value creation to separate util, cf. ed45a22
* Enables label editing for `bpmn:Group` elements  (via external label), cf. 0217a87
* Creates missing `bpmn:CategoryValue` when not existing, cf. a61609a

May-22-2019 08-58-22

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -13,6 +13,10 @@ import {
   isEventSubProcess
 } from '../util/DiUtil';
 
+import {
+  getLabel
+} from '../features/label-editing/LabelUtil';
+
 import { is } from '../util/ModelUtil';
 
 import {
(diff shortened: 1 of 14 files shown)


pinussilvestrus:
Should be ready for review now 👍

barmac:
Should the label stay in the same place when group is resized? It feels weird.

direct-editing

pinussilvestrus:
Good catch! I'll have a look at this.

Interesting: Groups are the first resizable elements which have an external label so we never had to deal with this scenario. But shouldn't be hard to fix.

pinussilvestrus:
After a discussion with @barmac it reveals that the label-move-on-resize case has lots of different scenarios to keep in mind of. We decided to open another issue for that and merge the changes here.

PR #1041, opened 2019-05-23
Allow to change intermediate events to boundary events

### LINKED ISSUE #478: Allow to change Intermediate Events to Boundary Events
* [x] Attach an intermediate event to a task to change it to a boundary event by dropping onto a task / sub-process border.
* [x] only works for some of catching and unspecified event definitions
* [ ] Keeps outgoing connections (if allowed)

barmac:
Closes #478

Builds on top of and includes changes in https://github.com/bpmn-io/bpmn-js/pull/1047

Requires https://github.com/bpmn-io/diagram-js/pull/354

### DIFF
--- a/lib/features/modeling/behavior/AttachEventBehavior.js
+++ b/lib/features/modeling/behavior/AttachEventBehavior.js
@@ -2,13 +2,14 @@ import inherits from 'inherits';
 
 import CommandInterceptor from 'diagram-js/lib/command/CommandInterceptor';
 
-import { is } from '../../../util/ModelUtil';
+import { isAny } from '../util/ModelingUtil';
+import { getBusinessObject } from '../../../util/ModelUtil';
 
 
 /**
  * BPMN specific attach event behavior
  */
-export default function AttachEventBehavior(eventBus, bpmnFactory, bpmnReplace) {
+export default function AttachEventBehavior(eventBus, bpmnReplace) {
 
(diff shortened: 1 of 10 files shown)


barmac:
Got stuck on layout:

When the event is replaced before the movement:
preExecute

When the event is replaced after the movement:
postExecuted

Goal:
Screenshot 2019-05-23 at 08 46 18

nikku:
:heart_eyes:

barmac:
This is not ready for (Parallel) Multiple Intermediate Catching Event 🙄

barmac:
@nikku SubProcess behaves correctly, because expansion conducted in `toggleCollapse` command triggers ResizeShapeHandler which... (you guessed it) lays out the connections. I'm going to restore the flag.

barmac:
@nikku Ready for review.

lib/util/model/ModelCloneUtils.js
```diff
@@ -9,6 +9,7 @@ import {
  * @type {Array}
  */
 export var IGNORED_PROPERTIES = [
+  'host',
```
nikku: What is this about?

barmac: I thought it might be useful in the context of host update, but actually we never use this. I'll remove it.

falko:
Thank you!

PR #1047, opened 2019-05-24
1045 boundary events are dettachable

### LINKED ISSUE #1045: Boundary Events are dettachable

It's frustrating that you have to copy event details and remove it if you want to turn a Boundary Event into Intermediate Event.

Boundary Events without event definition or with a definition which has a counterpart in Intermediate Events can be dettached and dropped on canvas as Intermediate Events. 

Do nothing and still disallow such action.

__Aditional context__

Not sure whether we want the Intermediate Event to be connected to the former parent. This could be a separate feature.

barmac:
Closes #1045

Requires https://github.com/bpmn-io/diagram-js/pull/354

### DIFF
--- a/lib/features/modeling/behavior/DetachEventBehavior.js
+++ b/lib/features/modeling/behavior/DetachEventBehavior.js
@@ -51,7 +51,7 @@ export default function DetachEventBehavior(eventBus, bpmnReplace) {
         };
       }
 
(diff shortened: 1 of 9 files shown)


barmac:
This will be failing until https://github.com/bpmn-io/diagram-js/pull/354 is merged.

PR #1052, opened 2019-05-28
fix(modeling): calculate appropriate new bounds on subprocess collapse

### LINKED ISSUE #1031: Constant position shifting of subprocess on repeated expand/collapse actions
Currently, expanding a collapsed subprocess with incoming sequence flows shifts the position of the subprocess to the right to [prevent overlap](https://github.com/camunda/camunda-modeler/issues/1244) with previous content.

If we collapse from expanded after this operation, the collapsed subprocess stays in the same position relative to the center of the expanded subprocess.

This means if we expand/collapse/expand/collapse/etc., the position of the element is constantly shifted towards the right.

bpmn-js#XXX

Is there a scenario where we would _not_ want to collapse back to the original position?

bpmn-js#XXX_02

gustavjf:

Closes #1031

bpmn-js#1031

Connection layout breaking is a [known issue](https://github.com/bpmn-io/bpmn-js/issues/1019).

### DIFF
--- a/lib/features/modeling/behavior/SubProcessBehavior.js
+++ b/lib/features/modeling/behavior/SubProcessBehavior.js
@@ -4,7 +4,10 @@ import CommandInterceptor from 'diagram-js/lib/command/CommandInterceptor';
 
 import { is } from '../../../util/ModelUtil';
 
-import { expandedBounds } from './ToggleElementCollapseBehaviour';
+import {
+  expandedBounds,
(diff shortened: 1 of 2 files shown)


nikku:
Closing this, as we decided to postpone further work on this issue.

The fix is squashed on this branch: https://github.com/bpmn-io/bpmn-js/pull/new/1244-expand-subprocess-right

PR #1062, opened 2019-06-04
Disable connection preview test for dragged new element

barmac:
This disables an integration test for connection preview when dragging a new element. The test will fail once https://github.com/bpmn-io/diagram-js/pull/355 is merged.


barmac:
@nikku As soon as we decide to display preview this test case can be re-enabled.

nikku:
Alright.

PR #1069, opened 2019-06-05
Move Label after Label Target Resizement

### LINKED ISSUE #1051: Group label should be moved after resizing

As a follow-up of https://github.com/bpmn-io/bpmn-js/pull/1040 it reveals that external labels of a `bpmn:Group` element should be moved after certain resize operations, e.g.

direct-editing

The label should be moved in a reasonable manner.

There are different scenarios where label movement after the resize operation makes sense.

cf. https://github.com/bpmn-io/bpmn-js/issues/1051#issuecomment-496390489

Leave it as it is.

__Environment__

 - Browser: Chrome 74
 - OS: macOS 10.14
 - Library version: current `master`

pinussilvestrus:
This automatically moves, if needed, an external label when the label target was resized. It uses a virtual `ReferencePoint` for every label and moves the label by the `ReferencePoints'` delta depending on the new bounds. 

Closes #1051 

Jun-05-2019 13-03-57

Jun-05-2019 13-06-00


pinussilvestrus:
Updated it accordingly:

* Move implementation to `LabelBehavior` to handle for _all_ external labels
* Use [`AttachUtil#getNewAttachPoint`](https://github.com/bpmn-io/diagram-js/blob/master/lib/util/AttachUtil.js#L20) to calculate new reference point delta

Just recognized that we already had the implementation for points 👍 Can use it without any modifications.

PR #1070, opened 2019-06-06
Correctly lay out boundary to activity loops

barmac:
Summary of changes:
* we no longer snap to target center during connection
* we snap inside of the target element
* we snap Boundary Event loops in the way to make sure the connection end is not too close
* layouter checks the area where cursor is placed to determine loop layout hints

Required by https://github.com/camunda/camunda-modeler/issues/903

See screenshots in the linked issue for the context.

### DIFF
--- a/lib/features/modeling/BpmnLayouter.js
+++ b/lib/features/modeling/BpmnLayouter.js
@@ -23,6 +23,8 @@ import {
 import { is } from '../../util/ModelUtil';
 
 
+var BOUNDARY_TO_HOST_THRESHOLD = 40;
+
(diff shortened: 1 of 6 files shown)


nikku:
We discussed this and agreed on looking into the connection behavior with sub-processes, too.

barmac:
It's not WIP anymore. @nikku can you please have a look at this once more?

Summary of changes:
* we no longer snap to target center during connection
* we snap inside of the target element
* we snap Boundary Event loops in the way to make sure the connection end is not too close
* layouter checks the area where cursor is placed to determine loop layout hints

PR #1094, opened 2019-06-19
feat(modeling): make initial host of Boundary Event sticky

### LINKED ISSUE #1078: Attached boundary events should be sticky

It is quite easy right now to accidentaly detach an attached boundary event:

bug

Boundary events should be sticky, i.e. snap to the boundary target and keep their attach overlay on move.

No valid alternatives.

barmac:
This PR makes Boundary Event's host initially sticky when moving the attachment. The purpose of this is to avoid accidental detachment. 

The hit box will shrink after the Boundary Event leaves the host and it will not grow back with next hover in order to prevent accidental attachment.

Closes #1078 

Requires https://github.com/bpmn-io/diagram-js/pull/368

### DIFF
--- a/lib/features/modeling/behavior/StickyBoundaryEventBehavior.js
+++ b/lib/features/modeling/behavior/StickyBoundaryEventBehavior.js
@@ -1,0 +1,61 @@
+import inherits from 'inherits';
+
+import CommandInterceptor from 'diagram-js/lib/command/CommandInterceptor';
+
+import {
+  classes as svgClasses
+} from 'tiny-svg';
+
+
+var EXTENDED_HIT = 'djs-hit-extended';
+
+/**
+ * BPMN specific detach event behavior
+ */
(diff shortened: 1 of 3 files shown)


lib/features/modeling/behavior/StickyBoundaryEventBehavior.js
```diff
@@ -0,0 +1,61 @@
...
+var EXTENDED_HIT = 'djs-hit-extended';
+
+/**
+ * BPMN specific detach event behavior
+ */
+export default function StickyBoundaryEventBehavior(eventBus, elementRegistry) {
+
+  CommandInterceptor.call(this, eventBus);
+
+  var hostGfx = null;
```
nikku: We should try to avoid caching stuff like that.

Remembering things in the drag context (event.context.***) is the preferred way to remember local state.

barmac:
Closing this as the solution will be provided via https://github.com/bpmn-io/diagram-js/pull/368

Issue #957, opened 2019-03-27
Improve navigation in container elements


pinussilvestrus:
__Is your feature request related to a problem? Please describe.__

As described in https://github.com/camunda/camunda-modeler/issues/238 it is very hard to navigate through huge container elements (as pools, lanes, subprocesses or, additionally, groups).

__Describe the solution you'd like__

Improve the situation. One possible solution is sketched in [this comment](https://github.com/camunda/camunda-modeler/issues/238#issuecomment-470062772):

> Applying to all container elements (pools, lanes and sub processes):
> 
> - container elements may only be moved after they have been selected
> - container elements may only be selected on the border (or when clicking their labels)
> - dragging inside a container element should move the canvas, if the container has not been selected

__Describe alternatives you've considered__

.

__Additional context__

* Relates to https://github.com/camunda/camunda-modeler/issues/238

barmac:
Reopened via https://github.com/bpmn-io/bpmn-js/pull/1118

nikku:
Restored on `develop` via https://github.com/bpmn-io/bpmn-js/commit/bee05f19a737b627c930e60fc991a3fab6fcd565.

PR #1097, opened 2019-06-20
Improve navigation inside large participants

barmac:
Related https://github.com/camunda/camunda-modeler/issues/238

Requires https://github.com/bpmn-io/diagram-js/pull/371

Closes https://github.com/bpmn-io/bpmn-js/issues/957

### DIFF
--- a/lib/Modeler.js
+++ b/lib/Modeler.js
@@ -24,6 +24,7 @@ import CreateModule from 'diagram-js/lib/features/create';
 import DistributeElementsModule from './features/distribute-elements';
 import EditorActionsModule from './features/editor-actions';
 import GridSnappingModule from './features/grid-snapping';
+import InteractionEventsModule from './features/interaction-events';
 import KeyboardModule from './features/keyboard';
 import KeyboardMoveSelectionModule from 'diagram-js/lib/features/keyboard-move-selection';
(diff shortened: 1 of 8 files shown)


nikku:
Does this cover sub-processes, too? We must include these, as they can be huge container elements, too.

barmac:
No, I will cover this then.

barmac:
So I tried the easy way and ended up with an infinite loop 😆 
loop

barmac:
Note for myself: I discovered that at some point SubProcess becomes a child of itself. Then, the recursive search destroys everything.

barmac:
Just pushed a working demo. It needs dynamic tests too, though. Notice change in diagram-js which occured to be required for the progress.

barmac:
This requires more work as we currently treat equally expanded and collapsed elements which should not take place.

barmac:
However, the ugly indefinite loops bugs are already resolved with recent commits :)

barmac:
@nikku I added collapsed shape support so this is ready for review.

barmac:
Apparently this seems to close https://github.com/bpmn-io/bpmn-js/issues/957

nikku:
> Apparently this seems to close #957

Apparently :wink:

Issue #1095, opened 2019-06-19
Can move attached boundary event onto task


nikku:
__Describe the Bug__

Moving an attached boundary event on on the host is possible right now, even in cases where it should not: 

![bug](https://user-images.githubusercontent.com/58601/59765787-1cabfe80-929f-11e9-8ae6-144c1883eb97.gif)

Actually dropping the element blows up the editor.

This is a regession. Stuff worked fine in bpmn-js@3.

__Steps to Reproduce__

* Model a task
* Attach a boundary event
* Move boundary event onto task middle
* Drop event

__Expected Behavior__

* Moving attached boundary events (detaching to intermediate events) is only possible inside `FlowElementContainers` (SubProcess, Participant / Process)

__Environment__

 - Browser: Any
 - OS: Any
 - Library version: `v4.0.0-beta.2`

nikku:
Closed via https://github.com/bpmn-io/bpmn-js/issues/1101.

PR #1101, opened 2019-06-24
Allow to drop Boundary Events only on containers

barmac:
Closes https://github.com/bpmn-io/bpmn-js/issues/1095


barmac:
@pinussilvestrus CI fails with nothing related to the PR :/
```
$ nvm install 10
curl: (7) Failed to connect to nodejs.org port 443: Connection timed out
Version '10' not found - try `nvm ls-remote` to browse available versions.
Failed to install 10. Remote repository may not be reachable.
```

PR #1109, opened 2019-06-27
Snap Bendpoint If Outside Of Target

barmac:
Closes https://github.com/bpmn-io/bpmn-js/issues/529


barmac:
The codecov/project check does not make any sense in this case :/

barmac:
Related https://github.com/bpmn-io/diagram-js/pull/372

Decided to close this until we know what the correct solution is.

barmac:
Wuffle fights this...

PR #1111, opened 2019-06-27
Restore import definitions in 3.x

barmac:
This PR restores `Viewer#importDefinitions` as it occurred that it is a useful and actively used method. I removed it in https://github.com/bpmn-io/bpmn-js/pull/983/commits/f8f933440484fbe97584d7fd8955b1d40e7c376e as it wasn't documented, tested or marked in any other way as part of public API. However, as its main functionality which is opening a diagram without XML import is extremely useful, it seems to be reasonable to actually add this to the Viewer's public API.


nikku:
> This PR restores `Viewer#importDefinitions` as it occurred that it is a useful and actively used method.

How do you deduce it is actively used?

nikku:
In any way, we [should use it ourselves](https://github.com/bpmn-io/bpmn-js/pull/1111/commits/0803df2d2e10c4241a41594ed5c57c46d3c2649d#diff-403c08b3642c7a9f03c30aae67d46a9aL215) if we decide to keep it around.

test/spec/ViewerSpec.js
```diff
@@ -742,8 +742,8 @@ describe('Viewer', function() {
           viewer.importXML(multipleXML);
 
           // then
-          viewer.on('import.done', function(event) {
-            done();
+          viewer.on('import.done', function(context) {
```
nikku: It is actually an event, not some context and I'd keep the _event_ identifier because of that.

test/spec/ViewerSpec.js
```diff
@@ -754,11 +754,11 @@ describe('Viewer', function() {
           var viewer = new Viewer({ container: container });
 
           // when
-          viewer.importXML(multipleXML, 'Diagram_2');
+          viewer.importXML(multipleXML, 'BpmnDiagram_2');
 
           // then
-          viewer.on('import.done', function(event) {
-            done();
+          viewer.on('import.done', function(context) {
```
nikku: It is actually an event, not some context and I'd keep the _event_ identifier because of that.

barmac:
Done.

nikku:
Merged via https://github.com/bpmn-io/bpmn-js/commit/63fabac89020318685f4650018d089ee67e6d2b8.

Please use different feature branches for backports and actual fix next time :tulip:

barmac:
OK, thanks for the feedback!

PR #1169, opened 2019-08-12
Allow to create attachments

barmac:
Related to https://github.com/camunda/camunda-modeler/issues/1462

### DIFF
--- a/lib/features/rules/BpmnRules.js
+++ b/lib/features/rules/BpmnRules.js
@@ -128,6 +128,10 @@ BpmnRules.prototype.init = function() {
         return canConnect(element.source, element.target, element);
       }
 
+      if (element.host) {
+        return canAttach(element, element.host, null, position);
+      }
+
       return canCreate(element, target, null, position);
     });
   });
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "diagram-js-direct-editing": "^1.6.0",
     "ids": "^1.0.0",
     "inherits": "^2.0.1",
-    "min-dash": "^3.5.0",
+    "min-dash": "^3.5.1",
     "min-dom": "^3.0.0",
     "object-refs": "^0.3.0",
     "tiny-svg": "^2.2.1"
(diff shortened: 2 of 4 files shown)


lib/features/rules/BpmnRules.js
```diff
@@ -128,6 +128,10 @@ BpmnRules.prototype.init = function() {
         return canConnect(element.source, element.target, element);
       }
 
+      if (element.host) {
+        return !!canAttach(element, element.host, null, position);
```
nikku: Can we drop the `!!` here?

test/spec/features/rules/BpmnRulesSpec.js
```diff
@@ -81,6 +81,36 @@ describe('features/modeling/rules - BpmnRules', function() {
...
+      var task = elementFactory.createShape({ type: 'bpmn:Task' }),
+          boundaryEvent = elementFactory.createShape({
+            type: 'bpmn:BoundaryEvent',
+            eventDefinitionType: 'bpmn:EscalationEventDefinition',
+            host: task
+          });
+
+      // then
+      expectCanCreate([ task, boundaryEvent ], 'Process', true);
+    }));
```
nikku: Add space.

lib/features/rules/BpmnRules.js
barmac: Yes, as soon as we merge https://github.com/bpmn-io/min-dash/pull/14 and update `min-dash`. Otherwise, we need to somehow handle the case that the rule might return `"attach"` or `true` based only on the selection order.

test/spec/features/rules/BpmnRulesSpec.js
barmac: Not sure where.

barmac:
> Can you update the message with a `rules` target:
> 
> ```
> fix(rules): allow ...
> ```

Done.

package.json
```diff
@@ -45,7 +45,7 @@
     "cpx": "^1.5.0",
     "del": "^3.0.0",
     "eslint": "^5.16.0",
-    "eslint-plugin-bpmn-io": "^0.8.2",
+    "eslint-plugin-bpmn-io": "github:bpmn-io/eslint-plugin-bpmn-io#5-require-empty-lines-before-comments",
```
nikku: Can we revert this change?

PR #1177, opened 2019-08-22
chore(project): remove coverage directory from the package

barmac:


nikku:
I believe this is a `npm` bug. Do we need to fix it in our code base?

nikku:
Apparently npm ignores what is part of `.gitignore`, even on `npm@6.10`.

PR #1186, opened 2019-09-07
fix: updates spelling mistakes - mostly just in comments

jonoirwinrsa:
_Which issue does this PR address?_

No issue open for it.

Closes #

Fixes some spelling mistakes - mostly in comments but once in a function name (only used in that module).

PR #1187, opened 2019-09-08
fix: updates spelling of "condition" in context provider

jonoirwinrsa:

None

Closes # None

Updates spelling of condtion to condition in context pad


pinussilvestrus:
Cf. [other pr comment](https://github.com/bpmn-io/bpmn-js/pull/1186#issuecomment-529343005) in which we want to update the commit message, in this case here it would be something like

```sh
chore(context-pad): update spelling
```

nikku:
Merged via https://github.com/bpmn-io/bpmn-js/commit/aea527775e06818efa34725728565c9e2c1a3f5f. Thanks!

PR #1216, opened 2019-10-07
fix(modeling): provide unique hash key for flow nodes when updating

oguzeroglu:
See [this comment](https://github.com/camunda/camunda-modeler/issues/1504#issuecomment-538992629)

Kapture 2019-10-07 at 15 43 43

Required by https://github.com/camunda/camunda-modeler/issues/1504

### DIFF
--- a/lib/features/modeling/cmd/UpdateFlowNodeRefsHandler.js
+++ b/lib/features/modeling/cmd/UpdateFlowNodeRefsHandler.js
@@ -34,7 +34,7 @@ UpdateFlowNodeRefsHandler.$inject = [
 
 UpdateFlowNodeRefsHandler.prototype.computeUpdates = function(flowNodeShapes, laneShapes) {
 
-  var handledNodes = {};
+  var handledNodes = [];
 
   var updates = [];
 
@@ -58,9 +58,9 @@ UpdateFlowNodeRefsHandler.prototype.computeUpdates = function(flowNodeShapes, la
   }
 
   function addFlowNodeShape(flowNodeShape) {
-    if (!handledNodes[flowNodeShape.id]) {
+    if (handledNodes.indexOf(flowNodeShape) == -1) {
       allFlowNodeShapes.push(flowNodeShape);
-      handledNodes[flowNodeShape.id] = flowNodeShape;
+      handledNodes.push(flowNodeShape);
     }
   }
 
@@ -92,7 +92,7 @@ UpdateFlowNodeRefsHandler.prototype.computeUpdates = function(flowNodeShapes, la
   laneShapes.forEach(function(laneShape) {
     var root = getLanesRoot(laneShape);
 
-    if (!root || handledNodes[root.id]) {
+    if (!root || handledNodes.indexOf(root) > -1) {
       return;
     }
 
@@ -102,7 +102,7 @@ UpdateFlowNodeRefsHandler.prototype.computeUpdates = function(flowNodeShapes, la
 
     children.forEach(addFlowNodeShape);
 
-    handledNodes[root.id] = root;
+    handledNodes.push(root);
   });
 
   flowNodeShapes.forEach(addFlowNodeShape);
--- a/test/spec/features/modeling/lanes/UpdateFlowNodeRefsSpec.js
+++ b/test/spec/features/modeling/lanes/UpdateFlowNodeRefsSpec.js
@@ -266,4 +266,28 @@ describe('features/modeling - lanes - flowNodeRefs', function() {
     expect(sourceLane.flowNodeRef).not.to.contain(event);
   }));
 
+
+  it('should not create duplicate refs on attaching / detaching', inject(function(elementRegistry, modeling) {
+
+    // given
+    var eventID = 'IntermediateThrowEvent',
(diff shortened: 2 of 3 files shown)


lib/features/modeling/cmd/UpdateFlowNodeRefsHandler.js
```diff
@@ -42,6 +42,10 @@ UpdateFlowNodeRefsHandler.prototype.computeUpdates = function(flowNodeShapes, la
 
   var allFlowNodeShapes = [];
 
+  function getNodeKey(node) {
```
nikku: This hack will only work on the assumption that the type always changes during replacement (which is not necessarily true). Did you consider an alternative solution approach that works without it?

oguzeroglu: The bug does not occur when I change elements using the replace pad (for example from one intermediate throw event to another one), it only occurs on Attach/Detach event behavior. So I assumed this would work. Is there a case where this would break? In any case I'll investigate why the bug does not occur on replace, but occurs on Attach/Detach behavior.

pinussilvestrus: The bug does not occur on replace because the newly created shape has another id temporarily on `preExecute` context. Maybe this would be an possibility also for detaching / attaching elements?

oguzeroglu: @nikku @pinussilvestrus Or, we can use Map here instead of Object for handledNodes variable. The problem with Objects that, their keys have to be String but Maps can have objects as key, so we can use directly the nodes as keys.

Check out this example:

```
var map = new Map();
var obj1 = {property1: true};
var obj2 = {property1: true};
map.set(obj1, "SOME VALUE");
map.has(obj2) // returns false
```
in this example obj1 and obj2 has exactly the same properties, values, but Map.has method returns false

pinussilvestrus: I think that's a good idea to ensure uniqueness. 👍 Nevertheless, it's interesting why we have the same id when detaching/attaching one element, and don't have the same behavior as in the replacement. We should maybe have a look at this in the future to prevent possible resulting bugs in other places, but for now, to fix this problem, your solution is good enough to fix the general uniqueness problem in the `UpdateFlowNodeRefsHandler` when we can't guarantee uniqueness with the ids.

oguzeroglu: I ended up finding a better solution and get rid of handledNodes control variable, as this variable is only used to check if we have processed an element or not, so actually allFlowNodeShapes.indexOf(node) gives us the same result anyway and it also provides uniqueness without going into id/type.

nikku: Did you think about the performance implications of not using a map? How many arrays are scanned how often for huge diagrams, i.e. [this one](https://demo.bpmn.io/s/application-processing).

lib/features/modeling/cmd/UpdateFlowNodeRefsHandler.js
```diff
@@ -101,8 +98,6 @@ UpdateFlowNodeRefsHandler.prototype.computeUpdates = function(flowNodeShapes, la
     });
 
     children.forEach(addFlowNodeShape);
-
-    handledNodes[root.id] = root;
```
pinussilvestrus: Can we remove this without any replacement? It will iterate over the same children again for two lanes within the same participant.

lib/features/modeling/cmd/UpdateFlowNodeRefsHandler.js
oguzeroglu: No I did not consider this, however we cannot use Map as they are not compatible with the JS version used in our test framework (PhantomJS).

As a performance remark:

```
var ary = [];
for (var i = 0; i < 1000000; i++){
	ary.push({testProp: true});
}
var obj = {testProp: true};
ary.push(obj); // put array to last as the worst case scenario

var startTime = performance.now();
ary.indexOf(obj);
console.log(performance.now() - startTime);  // 0.8 ms
```

So it takes 0.8 ms for indexOf to execute for 1 million + 1 nodes. I believe that amount of nodes are rarely the case and we have another problems caused by usage of document.elementFromPoint in Hover related classes if we happen to have a diagram with 1M+ nodes.

The comparison of objects are performed with (obj1 === obj), which is basically a reference comparison, so it's rather cheap.

nikku: Thanks for the clarification. I did another test and verified we don't run in a `O^n` setting here.

lib/features/modeling/cmd/UpdateFlowNodeRefsHandler.js
oguzeroglu: Actually it should not add the same children as the addFlowNodeShape does an indexOf check and also at the beginning of the method we do an early return also if the allFlowNodeShapes contain root. Does that make sense to you? Or maybe I misunderstood your point?

PR #1218, opened 2019-10-09
feat(label-editing): complete editing on editor save

oguzeroglu:
Related to: https://github.com/camunda/camunda-modeler/issues/1473

Also see: https://github.com/camunda/camunda-modeler/pull/1527

### DIFF
--- a/lib/features/editor-actions/BpmnEditorActions.js
+++ b/lib/features/editor-actions/BpmnEditorActions.js
@@ -51,6 +51,7 @@ BpmnEditorActions.prototype._registerDefaultActions = function(injector) {
   var directEditing = injector.get('directEditing', false);
   var searchPad = injector.get('searchPad', false);
   var modeling = injector.get('modeling', false);
+  var eventBus = injector.get('eventBus', false);
 
   // (2) check components and register actions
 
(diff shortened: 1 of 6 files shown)


lib/features/editor-actions/BpmnEditorActions.js
```diff
@@ -174,4 +175,10 @@ BpmnEditorActions.prototype._registerDefaultActions = function(injector) {
     });
   }
 
-};
\ No newline at end of file
+  if (eventBus) {
+    this._registerAction('saveEditing', function() {
```
pinussilvestrus: Can't we directly use `directEditing#complete` here? Is there a reason why we have to introduce a dedicated event for this?

oguzeroglu: Because the completion if direct editing is taken care by LabelEditingProvider class [here](https://github.com/bpmn-io/bpmn-js/blob/44ba918d7e823001e1e2f7aebff44022ae518f5e/lib/features/label-editing/LabelEditingProvider.js#L55), rather than duplicating the direct editing completion logic, I wanted LabelEditingProvider to take care of this case also

oguzeroglu:
I'm closing this PR in order to implement this feature under tabs/ in camunda modeler, so we won't have to do 3 PRs for 3 different projects.

PR #1227, opened 2019-10-23
Skip label moving if label is not yet created.

oguzeroglu:
Related to https://github.com/zeebe-io/zeebe-modeler/issues/134

### DIFF
--- a/lib/features/modeling/behavior/LabelBehavior.js
+++ b/lib/features/modeling/behavior/LabelBehavior.js
@@ -212,7 +212,12 @@ export default function LabelBehavior(
     var label = event.context.connection.label,
         labelAdjustment;
 
-    if (!label) {
+    // handle missing label as well as the case
+    // that the label parent does not exist (yet),
+    // because it is being pasted / created via multi element create
+    //
+    // Cf. https://github.com/bpmn-io/bpmn-js/pull/1227
+    if (!label || !label.parent) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


lib/features/modeling/behavior/LabelBehavior.js
```diff
@@ -212,7 +212,7 @@ export default function LabelBehavior(
     var label = event.context.connection.label,
         labelAdjustment;
 
-    if (!label) {
+    if (!label || !label.parent) {
```
oguzeroglu: I find an early return here is to be the best solution. The reason for it is that the only case where `connection.updateWaypoints` is executed for a flow having a not-yet-created label is during the copy-paste operation. Since the label will be created after this block, as a part of `elements.create` exactly in the right position, we simply don't need to move it here.

nikku: Makes sense. Could you add an additional comment regarding this that makes clear which scenarios we're covering here? That helps us to understand what is going on here and why this case is excluded next time we look into the issue.

oguzeroglu: The problem occurs when LayoutConnectionBehavior of GridSnapping module intervenes during elements.create in order to snap waypoints of SequenceFlows. This only happens for SequenceFlows having waypoints.length > 2. This behavior triggers updateWaypoints which is later catched by LabelBehavior in order to correct the label position which is attached to a SequenceFlow. However since the label is not yet created here, since this all happens during elements.create, the operation fails.

nikku:
Thanks for updating this PR (adding an integration test).

I did a small amount of minor updates on top that I'd like you to inspect (cf. https://github.com/bpmn-io/bpmn-js/commit/b98dbca45685b1fc525078e3e4e080ed6322332a):

* I documented the `!label.parent` check as it can easily look like a bad coding practice otherwise
* I've pulled out that specific test (+ dependency to grid snapping) into its own test suite, using its own  diagram
* I've simplified the test case, asserting an absolute, expected position. Otherwise it's can be hard to assess what's actually being verified.

PR #1232, opened 2019-11-06
chore(project): use source maps in tests

barmac:
Source maps are really helpful when debugging and don't significantly change the bundle time.


nikku:
Source maps were a pain in the back in the past. I'm happy to give them another try.

Let's keep a close eye on whether their helpfulness actually outweighs the issues they bring.

PR #1234, opened 2019-11-08
test(popup-menu): remove dependency on internals

barmac:
This fixes the tests that depended on diagram-js Popup Menu internals. 

Requires https://github.com/bpmn-io/diagram-js/pull/431

### DIFF
--- a/test/spec/features/popup-menu/ReplaceMenuProviderSpec.js
+++ b/test/spec/features/popup-menu/ReplaceMenuProviderSpec.js
@@ -1884,7 +1884,9 @@ describe('features/popup-menu - replace menu provider', function() {
 
 // helper /////
 function queryEntry(id) {
-  return domQuery('.djs-popup [data-id="' + id + '"]');
+  var container = getBpmnJS().get('canvas').getContainer();
+
+  return domQuery('.djs-popup [data-id="' + id + '"]', container);
 }
 
 /**
@@ -1895,7 +1897,9 @@ function queryEntry(id) {
  * @return {<Array>}
  */
 function queryEntries() {
-  return domQueryAll('.djs-popup .entry');
+  var container = getBpmnJS().get('canvas').getContainer();
+
+  return domQueryAll('.djs-popup .entry', container);
 }
 
 function triggerAction(popupMenu, id) {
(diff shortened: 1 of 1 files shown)


test/spec/features/popup-menu/ReplaceMenuProviderSpec.js
```diff
@@ -1991,3 +1879,31 @@ describe('features/popup-menu - replace menu provider', function() {
   });
 
 });
+
+
+
+// helper /////
+function queryEntry(id) {
+  return domQuery('.djs-popup [data-id="' + id + '"]');
```
nikku: I believe we must scope this one to the current diagram-js instance, right? This is what we accomplished by passing the popupMenu in the past. Otherwise we assert that the entry from the first matching test run is there, not the current one.

barmac: Yes and no. 

**Yes**, because at the moment we query the whole document which surely slows tests down. 

**No**, because there cannot be a second popup menu open since the old diagram is [always destroyed before a next one is created](https://github.com/bpmn-io/bpmn-js/blob/33156e43be92b57bdef0f3b7e965ce36b105fe1f/test/helper/index.js#L117). This is why the tests didn't break.

Anyway, I pushed a scoped container queries on top because it's surely not a good practice to query outside of the test container.

test/spec/features/popup-menu/ReplaceMenuProviderSpec.js
```diff
@@ -658,8 +667,8 @@ describe('features/popup-menu - replace menu provider', function() {
           openPopup(startEvent);
 
           // then
-          expect(queryEntry('replace-with-none-start')).to.be.null;
-          expect(queryEntries()).to.have.length(6);
+          expect(queryEntry('replace-with-none-start'), container).to.be.null;
```
nikku: Wrong position of container argument.

nikku: We can get rid of the container argument entirely:

```javascript
function queryEntry(selector) {
  const container = getBpmnJS().get('canvas')._container;

  ...
}
```

barmac: Fixed.

nikku:
Good stuff :cake:.

PR #1240, opened 2019-11-18
chore(ci): don't install diagram-js@develop if target is master

barmac:
This should prevent `bpmn-js` fix branches from failing when there are breaking changes pending on `diagram-js@develop`.

Related to https://github.com/bpmn-io/bpmn-js/pull/1239#event-2803052610


barmac:
Even if we merge this, `push` builds won't stop to fail.

Issue #258, opened 2015-05-02
Viewer / Modeler modules should be replacable


adrobisch:
For the initialization of a Modeler or Viewer, its currently only possible to define either the whole set of modules or some additional modules via the options object. The latter is not sufficient to change default behavior and the former requires to much knowledge about internals.

There should be a way to replace modules by name. e.g. 

```javascript
var options = {
  replaceModules: {
    "contextPadProvider": {...} 
  }
  ...
}
```

nikku:
You can override diagram components by simply redefining them in override modules, i.e. modules that provide a interface compatible version of the service with the same name.

I added some documentation to the viewer and modeler entry points mentioning this option.

nikku:
See [relevant docs section](https://github.com/bpmn-io/bpmn-js/commit/6b23d181d5db9fa2dd07500208b46080a4d4dd9b#diff-6a28cb13662494293351ba71b07c2a92R68)

PR #1252, opened 2019-12-01
feat(project): provide Base{Viewer|Modeler} distributions

nikku:
This allows users to use the viewer / modeler features without inheriting our modules.

Related to #258


lib/BaseModeler.js
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * Have a look at {@link Modeler} for a bundle that includes actual features.
+ *
+ * @param {Object} [options] configuration options to pass to the viewer
+ * @param {DOMElement} [options.container] the container to render the viewer in, defaults to body.
+ * @param {String|Number} [options.width] the width of the viewer
+ * @param {String|Number} [options.height] the height of the viewer
+ * @param {Object} [options.moddleExtensions] extension packages to provide
+ * @param {Array<didi.Module>} [options.modules] a list of modules to override the default modules
+ * @param {Array<didi.Module>} [options.additionalModules] a list of modules to use with the default modules
```
oguzeroglu: I don't quite get this parameter in BaseModeler. Now that the _modules is empty for BaseModeler/BaseViewer, why would the user pass an additinalModules parameter? Maybe I'm missing a point here though.

nikku: One could argue that we could get rid of this one.

I did not change and APIs, so that `BaseModeler` and `Modeler` remain interface equivalent.

PR #1272, opened 2020-01-31
Copy message referenced by receive and send task

barmac:
Related to https://github.com/camunda/camunda-modeler/issues/1637


nikku:
What about send tasks :lollipop: ?

barmac:
> What about send tasks 🍭 ?

I checked that before in Camunda Modeler and it seems that we don't allow `messageRef` for send task.

However, I checked now the spec and it makes perfect sense to add the feature for send tasks too.

barmac:
@pinussilvestrus ready

PR #1276, opened 2020-02-06
feat(modeling): create generic ids for new elements

barmac:
This makes sure that the semantic ID prefix reflects
the common type of all replace options.

ID prefixes for elements will match the examples:
  * `bpmn:ServiceTask` => `Activity_<id_suffix>`
  * `bpmn:EndEvent` => `Event_<id_suffix>`
  * `bpmn:EventBasedGateway` => `Gateway_<id_suffix>`
  * `bpmn:SequenceFlow` => `Flow_<id_suffix>`

Related to https://github.com/camunda/camunda-modeler/issues/1654

---

Why `bpmn:ServiceTask` => `Activity_<id_suffix>`? Because we allow to morph tasks into subprocesses and call activities.

### DIFF
--- a/lib/features/modeling/BpmnFactory.js
+++ b/lib/features/modeling/BpmnFactory.js
@@ -8,6 +8,10 @@ import {
   isAny
 } from './util/ModelingUtil';
 
+import {
+  is
+} from '../../util/ModelUtil';
+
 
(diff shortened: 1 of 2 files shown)


nikku:
This should solve more than it breaks, let's merge it.

:heart:

PR #1295, opened 2020-03-20
fix(modeling): re-layout participant's connection when lane added

### LINKED ISSUE #1289: Adding lane does not re-connect message flows
__Describe the Bug__

When adding a lane _above_ or _below_, connected message flows are not re-connected:

capture AjhP9h_optimized

__Steps to Reproduce__

* Open [test diagram](https://github.com/bpmn-io/bpmn-js/files/4306597/collaboration-message-flows.bpmn.txt)
* Add lane below upper participant
* __Observe message flow not re-connecting__

__Expected Behavior__

Message flow re-connects.

__Environment__

 - Browser: Any
 - OS: Any
 - Library version: v6.3.1

barmac:
Closes #1289

### DIFF
--- a/lib/features/modeling/cmd/AddLaneHandler.js
+++ b/lib/features/modeling/cmd/AddLaneHandler.js
@@ -1,5 +1,6 @@
 import {
-  filter
+  filter,
+  forEach
 } from 'min-dash';
 
 import {
@@ -98,4 +99,18 @@ AddLaneHandler.prototype.preExecute = function(context) {
     width: shape.width - (isRoot ? LANE_INDENTATION : 0),
     height: 120
   }, laneParent);
+};
+
+AddLaneHandler.prototype.postExecute = function(context) {
+
+  // (3) Re-layout participant's connections.
+  var modeling = this._modeling,
+      shape = context.shape,
+      participant = getLanesRoot(shape),
+      connections = participant.incoming.concat(participant.outgoing);
+
+
+  forEach(connections, function(c) {
+    modeling.layoutConnection(c);
+  });
 };
--- a/test/spec/features/modeling/lanes/AddLaneSpec.js
+++ b/test/spec/features/modeling/lanes/AddLaneSpec.js
@@ -5,6 +5,7 @@ import {
 } from 'test/TestHelper';
 
 import {
+  map,
   pick
 } from 'min-dash';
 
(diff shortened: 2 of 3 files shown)


lib/features/modeling/cmd/AddLaneHandler.js
```diff
@@ -99,3 +100,17 @@ AddLaneHandler.prototype.preExecute = function(context) {
     height: 120
   }, laneParent);
 };
+
+AddLaneHandler.prototype.postExecute = function(context) {
```
nikku: Why is that not happening automatically, as the participant is resized during lane add? Maybe we need to change the resizing / moving shapes that are affected by the resize operation?

barmac: The reason is that we disabled re-layouting when using space tool: https://github.com/bpmn-io/diagram-js/blob/aeb6670c09d8e93f25335148656b1754450e44bf/lib/features/modeling/cmd/SpaceToolHandler.js#L67 (via https://github.com/bpmn-io/diagram-js/pull/364). I will look into that if we can just avoid using space tool so that we don't need to re-layout manually.

barmac: So alternatively we could unbuild the dependency on the space tool and introduce the following flow:
1) resize participant => grow it in the direction depending on position of the added lane
2) move lanes and lanes' children
3) create the actualn lane.

1 & 2 are basically what the space tool does. The only difference will be that we start to rely on automatic layouting of the resize shape handler which was disabled for space tool.

What do you think?

lib/features/modeling/cmd/AddLaneHandler.js
```diff
@@ -99,3 +100,17 @@ AddLaneHandler.prototype.preExecute = function(context) {
...
+
+  // (3) Re-layout participant's connections.
+  var modeling = this._modeling,
+      shape = context.shape,
+      participant = getLanesRoot(shape),
+      connections = participant.incoming.concat(participant.outgoing);
+
+
+  forEach(connections, function(c) {
+    modeling.layoutConnection(c);
```
nikku: Source changed / target changed hints may be relevant here. Did you check that?

barmac: First it seemed unnecessary, but then I just discovered an issue with participants placed next to the resized one. I will have a fix for that.

lib/features/modeling/cmd/AddLaneHandler.js
nikku: > So alternatively we could unbuild the dependency on the space tool and introduce the following flow.

This is exactly what the space tool does. In my opinion it makes little sense to duplicate it.

I guess your solution is fine for the time being --- as you mentioned we've unbuilt all layouting from the space tool with the last release. As a result we need to compensate for the lack thereof in the places we forgot.

barmac:
Closing this in favor of an upcoming solution in diagram-js repo.

Issue #1296, opened 2020-03-26
Link Events are missing the `name` attribute


RomanJRW:
__Describe the Bug__

When Link events (either catch or throw) are added to a model, they do not get given a default `name` attribute. I believe that this is a required attribute according to specification, so the resulting model is not considered valid. To demonstrate the failed validation, if I try to deploy a diagram containing link events that haven't been explicitly given a name from the Camunda Modeler, I get the following error:

```
Deploy error: {"deploymentName":"diagram_5","message":"ENGINE-09005 Could not parse BPMN process. Errors: \n* cvc-complex-type.4: Attribute 'name' must appear on element 'bpmn:linkEventDefinition'. | resource diagram_5.bpmn | line 13 | column 35\n* cvc-complex-type.4: Attribute 'name' must appear on element 'bpmn:linkEventDefinition'. | resource diagram_5.bpmn | line 18 | column 35","status":400,"statusText":"Bad Request","url":"http://localhost:8080/engine-rest/deployment/create"} [ deploy-error ]
```
I originally found this in Optimize where we are also getting the same validation error, but it's easier to reproduce in Camunda Modeler.

__Steps to Reproduce__

1. Create a model containing either or both of link throw intermediate event and link catch intermediate event. Do not set the name property manually.
2. Validate process model, for example when deploying in Camunda Modeler

__Expected Behavior__

I'm not totally sure what the preferred behaviour would be. One option would be for a default value to be assigned to the `name` attribute, perhaps using the id of the intermediate event that it is part of. Although this wouldn't necessarily be useful until overridden, it would at least still not leave models in a potentially invalid state.

__Environment__

 - Browser: Chrome 80.0.3987.149
 - OS: MacOS 10.14.5
 - Library version: 6.3.4

nikku:
Thanks for reporting this issue!

Do you have a valid diagram that can be deployed (and executed) on Camunda?

RomanJRW:
No worries 🙂 

I've had to attach as .txt files as GitHub doesn't allow bpmn or xml files. As you can see, the only difference between the deployable diagram and the undeployable diagram is the presence of the `name` attribute on the link event, which matches the logged parsing exception message too.

[deployable_links.txt](https://github.com/bpmn-io/bpmn-js/files/4388699/deployable_links.txt)
[undeployable_links.txt](https://github.com/bpmn-io/bpmn-js/files/4388700/undeployable_links.txt)

nikku:
[...]

PR #1297, opened 2020-03-26
feat(replace): create bpmn:LinkEventDefinition with empty name

nikku:
Closes https://github.com/bpmn-io/bpmn-js/issues/1296


nikku:
Thanks for the better test cases verifying the behavior!

Ready for round :two: of review.

PR #1298, opened 2020-03-27
Improve layouting of boundary elements and connections on host/source/target resize

barmac:
New behavior: If the attachment or connections can stay in same place after resize, keep it there.

Related to https://github.com/bpmn-io/bpmn-js/issues/1289
Related to https://github.com/camunda/camunda-modeler/issues/1737
Related to https://github.com/camunda/camunda-modeler/issues/655

Requires https://github.com/bpmn-io/diagram-js/pull/452


barmac:
sticky

Maybe it should be even more sticky. Notice how the event is moved when both bottom and right edge are changed. What do you think?

barmac:
sticky-connection

Sticky connection for comparison.

nikku:
I do not see a difference :thinking:.

barmac:
> I do not see a difference 🤔.

Then I guess user won't see it either.

philippfromme:
I can see the difference. And yes, it would be nice to have.

barmac:
OK, I'll try to implement it.

philippfromme:
@barmac Let me know when this is ready to be reviewed. 👍

barmac:
@philippfromme @nikku Ready for review.

PR #1301, opened 2020-03-30
Public apis promisified

### LINKED ISSUE #812: Switch from callbacks to promises
We should use promises rather than callbacks in parts of our API that require asynchronity:

* `Viewer#importXML`
* `Viewer#saveXML`
* `Viewer#saveSVG`
* `Moddle#fromXML`
* `Moddle#toXML`

Why switch:

* interact nicely with ES6 `async/await`
* easier consumption for the user

__Tasks__

* [x] Decide for transition strategy
* [x] Roll out change (https://github.com/bpmn-io/bpmn-js/issues/1311)
* [ ] Update documentation (https://github.com/bpmn-io/bpmn-js-examples/issues/112)
* [ ] Blog about it (https://github.com/bpmn-io/bpmn-js/issues/1320)
* [x] :heart: IE11

oguzeroglu:
Closes #812

Ticket for moddle APIs: https://github.com/bpmn-io/bpmn-moddle/issues/71

### DIFF
--- a/lib/util/PromisifierUtil.js
+++ b/lib/util/PromisifierUtil.js
@@ -1,0 +1,61 @@
+import { isFunction } from 'min-dash';
+
+export function promisifyIfNecessary(api) {
+
+  return function() {
+
+    var argLen = arguments.length;
+    var hasCallback = argLen > 0 && isFunction(arguments[argLen - 1]);
+
+    if (hasCallback) {
+      console.warn(
+        'Warning: passing callbacks to ' + api.name +
+        ' API will be deprecated in the next major bpmn-js release.' +
+        ' Consider switching to promises instead. See the API documentation.'
+      );
+
+      return api.apply(this, arguments);
+    }
+
+    if (!window.Promise) {
+      throw new Error('Promises are not supported for this browser. Consider polyfilling Promises.');
+    }
+
+    return new Promise(function(resolve, reject) {
+
+      var callback = generateCallbackFunction(resolve, reject);
+      var newArguments = concatFunctionArguments(this.originalArguments, callback);
+
+      api.apply(this.originalContext, newArguments);
+    }.bind({ originalContext: this, originalArguments: arguments }));
+  };
+}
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -39,6 +39,7 @@
     "*.css"
   ],
   "devDependencies": {
+    "babel-polyfill": "^6.26.0",
```
nikku: Can we only Polyfill `Promise` please? We do not want to blindly polyfill any current and future ES improvement (yet).

lib/util/PromisifierUtil.js
```diff
@@ -0,0 +1,66 @@
+import { isFunction, some } from 'min-dash';
+
+export function promisifyIfNecessary(api) {
+
+  return function() {
+
+    // @oguz:
```
nikku: Isn't the function always the last argument? At least that is common convention in callback style APIs.

lib/util/PromisifierUtil.js
```diff
@@ -0,0 +1,66 @@
...
+      );
+
+      return api.apply(this, arguments);
+    }
+
+    if (!window.Promise) {
+      throw new Error('Promises are not supported for this browser. Consider polyfilling Promises.');
+    }
+
+    // eslint-disable-next-line no-undef
```
nikku: Declare a global `Promise` variable in eslint to get rid of the warning instead.

test/spec/ViewerSpec.js
```diff
@@ -798,9 +800,15 @@ describe('Viewer', function() {
         try {
           viewer.importXML(xml);
```
nikku: Must await this.

oguzeroglu: I don't get why we want to await this. In this test we want to see if viewer.importXML throws an exception if we don't pass a callback and Promises are not polyfilled.

oguzeroglu:
We'll change our approach on this topic.

PR #1311, opened 2020-04-17
APIs promisified

### LINKED ISSUE #812: Switch from callbacks to promises

oguzeroglu:
This PR promisifies the following APIs:

BaseViewer#importXML
BaseViewer#importDefinitions
BaseViewer#open
BaseViewer#saveXML
BaseViewer#saveSVG
Modeler#createDiagram

Tests are also promisifed.

We should have a documentation here: https://bpmn.io/l/moving-to-promises.html

----

Child of #812

### DIFF
--- a/lib/BaseViewer.js
+++ b/lib/BaseViewer.js
@@ -104,25 +104,25 @@ BaseViewer.prototype.importXML = function(xml, bpmnDiagram, done) {
   // allow xml manipulation
   xml = this._emit('import.parse.start', { xml: xml }) || xml;
 
-  this._moddle.fromXML(xml, 'bpmn:Definitions', function(err, definitions, context) {
+  this._moddle.fromXML(xml, 'bpmn:Definitions').then(function(result) {
+    var definitions = result.rootElement;
+    var references = result.references;
+    var parseWarnings = result.warnings;
+    var elementsById = result.elementsById;
+
+    var context = {
+      elementsById: elementsById,
+      references: references,
+      warnings: parseWarnings
+    };
 
     // hook in post parse listeners +
     // allow definitions manipulation
     definitions = self._emit('import.parse.complete', {
-      error: err,
+      error: null,
       definitions: definitions,
       context: context
     }) || definitions;
-
-    var parseWarnings = context.warnings;
-
-    if (err) {
-      err = checkValidationError(err);
-
-      self._emit('import.done', { error: err, warnings: parseWarnings });
-
-      return done(err, parseWarnings);
-    }
 
     self.importDefinitions(definitions, bpmnDiagram, function(err, importWarnings) {
       var allWarnings = [].concat(parseWarnings, importWarnings || []);
@@ -131,6 +131,17 @@ BaseViewer.prototype.importXML = function(xml, bpmnDiagram, done) {
 
       done(err, allWarnings);
     });
+  }).catch(function(err) {
+
+    self._emit('import.parse.complete', {
+      error: err
+    });
+
+    err = checkValidationError(err);
+
+    self._emit('import.done', { error: err, warnings: err.warnings });
+
+    return done(err, err.warnings);
   });
 };
 
@@ -263,23 +274,27 @@ BaseViewer.prototype.saveXML = function(options, done) {
     definitions: definitions
   }) || definitions;
 
-  this._moddle.toXML(definitions, options, function(err, xml) {
+  this._moddle.toXML(definitions, options).then(function(result) {
+
+    var xml = result.xml;
 
     try {
       xml = self._emit('saveXML.serialized', {
-        error: err,
+        error: null,
         xml: xml
       }) || xml;
 
       self._emit('saveXML.done', {
-        error: err,
+        error: null,
         xml: xml
       });
     } catch (e) {
       console.error('error in saveXML life-cycle listener', e);
     }
 
-    done(err, xml);
+    done(null, xml);
+  }).catch(function(err) {
+    done(err);
   });
 };
 
--- a/lib/BaseViewer.js
+++ b/lib/BaseViewer.js
@@ -7,7 +7,6 @@
 import {
   assign,
   find,
-  isFunction,
   isNumber,
   omit
 } from 'min-dash';
@@ -68,6 +67,21 @@ export default function BaseViewer(options) {
 
 inherits(BaseViewer, Diagram);
 
+/**
+* The importXML result.
+*
+* @typedef {Object} ImportXMLResult
+*
+* @property {Array<string>} warnings
+*/
+
+/**
+* The importXML error.
+*
+* @typedef {Error} ImportXMLError
+*
+* @property {Array<string>} warnings
+*/
 
 /**
  * Parse and render a BPMN 2.0 diagram.
@@ -89,64 +103,82 @@ inherits(BaseViewer, Diagram);
  *
  * @param {string} xml the BPMN 2.0 xml
  * @param {ModdleElement<BPMNDiagram>|string} [bpmnDiagram] BPMN diagram or id of diagram to render (if not provided, the first one will be rendered)
- * @param {Function} [done] invoked with (err, warnings=[])
+ *
+ * Returns {Promise<ImportXMLResult, ImportXMLError>}
  */
-BaseViewer.prototype.importXML = wrapForCompatibility(function importXML(xml, bpmnDiagram, done) {
-
-  if (isFunction(bpmnDiagram)) {
-    done = bpmnDiagram;
-    bpmnDiagram = null;
-  }
-
-  // done is optional
-  done = done || function() {};
+BaseViewer.prototype.importXML = wrapForCompatibility(function importXML(xml, bpmnDiagram) {
 
   var self = this;
 
-  // hook in pre-parse listeners +
-  // allow xml manipulation
-  xml = this._emit('import.parse.start', { xml: xml }) || xml;
+  return new Promise(function(resolve, reject) {
 
-  this._moddle.fromXML(xml, 'bpmn:Definitions').then(function(result) {
-    var definitions = result.rootElement;
-    var references = result.references;
-    var parseWarnings = result.warnings;
-    var elementsById = result.elementsById;
+    // hook in pre-parse listeners +
+    // allow xml manipulation
+    xml = self._emit('import.parse.start', { xml: xml }) || xml;
 
-    var context = {
-      elementsById: elementsById,
-      references: references,
-      warnings: parseWarnings
-    };
+    self._moddle.fromXML(xml, 'bpmn:Definitions').then(function(result) {
+      var definitions = result.rootElement;
+      var references = result.references;
+      var parseWarnings = result.warnings;
+      var elementsById = result.elementsById;
 
-    // hook in post parse listeners +
-    // allow definitions manipulation
-    definitions = self._emit('import.parse.complete', {
-      error: null,
-      definitions: definitions,
-      context: context
-    }) || definitions;
+      var context = {
+        elementsById: elementsById,
+        references: references,
+        warnings: parseWarnings
+      };
 
-    self.importDefinitions(definitions, bpmnDiagram, function(err, importWarnings) {
-      var allWarnings = [].concat(parseWarnings, importWarnings || []);
+      // hook in post parse listeners +
+      // allow definitions manipulation
+      definitions = self._emit('import.parse.complete', {
+        error: null,
+        definitions: definitions,
+        context: context
+      }) || definitions;
 
-      self._emit('import.done', { error: err, warnings: allWarnings });
+      self.importDefinitions(definitions, bpmnDiagram).then(function(result) {
+        var allWarnings = [].concat(parseWarnings, result.warnings || []);
 
-      done(err, allWarnings);
+        self._emit('import.done', { error: null, warnings: allWarnings });
+
+        return resolve({ warnings: allWarnings });
+      }).catch(function(err) {
+        var allWarnings = [].concat(parseWarnings, err.warnings || []);
+
+        self._emit('import.done', { error: err, warnings: allWarnings });
+
+        return reject(addWarningsToError(err, allWarnings));
+      });
+    }).catch(function(err) {
+
+      self._emit('import.parse.complete', {
+        error: err
+      });
+
+      err = checkValidationError(err);
+
+      self._emit('import.done', { error: err, warnings: err.warnings });
+
+      return reject(err);
     });
-  }).catch(function(err) {
-
-    self._emit('import.parse.complete', {
-      error: err
-    });
-
-    err = checkValidationError(err);
-
-    self._emit('import.done', { error: err, warnings: err.warnings });
-
-    return done(err, err.warnings);
   });
 });
+
+/**
+* The importDefinitions result.
+*
+* @typedef {Object} ImportDefinitionsResult
+*
+* @property {Array<string>} warnings
+*/
+
+/**
+* The importDefinitions error.
+*
+* @typedef {Error} ImportDefinitionsError
+*
+* @property {Array<string>} warnings
+*/
 
 /**
  * Import parsed definitions and render a BPMN 2.0 diagram.
@@ -165,22 +197,44 @@ BaseViewer.prototype.importXML = wrapForCompatibility(function importXML(xml, bp
  *
  * @param {ModdleElement<Definitions>} definitions parsed BPMN 2.0 definitions
  * @param {ModdleElement<BPMNDiagram>|string} [bpmnDiagram] BPMN diagram or id of diagram to render (if not provided, the first one will be rendered)
- * @param {Function} [done] invoked with (err, warnings=[])
+ *
+ * Returns {Promise<ImportDefinitionsResult, ImportDefinitionsError>}
  */
-BaseViewer.prototype.importDefinitions = wrapForCompatibility(function importDefinitions(definitions, bpmnDiagram, done) {
+BaseViewer.prototype.importDefinitions = wrapForCompatibility(function importDefinitions(definitions, bpmnDiagram) {
 
-  if (isFunction(bpmnDiagram)) {
-    done = bpmnDiagram;
-    bpmnDiagram = null;
-  }
+  var self = this;
 
-  // done is optional
-  done = done || function() {};
+  return new Promise(function(resolve, reject) {
 
-  this._setDefinitions(definitions);
+    self._setDefinitions(definitions);
 
-  return this.open(bpmnDiagram, done);
+    self.open(bpmnDiagram).then(function(result) {
+
+      var warnings = result.warnings;
+
+      return resolve({ warnings: warnings });
+    }).catch(function(err) {
+
+      return reject(err);
+    });
+  });
 });
+
+/**
+ * The open result.
+ *
+ * @typedef {Object} OpenResult
+ *
+ * @property {Array<string>} warnings
+ */
+
+/**
+* The open error.
+*
+* @typedef {Error} OpenError
+*
+* @property {Array<string>} warnings
+*/
 
 /**
  * Open diagram of previously imported XML.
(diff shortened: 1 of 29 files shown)


oguzeroglu:
IMO codecov check can be ignored. The only thing we can do is to check if certain APIs are promisified/do a console.warn if we pass callbacks/throw error if Promise is not defined. These are not useful tests IMO. Any other existing tests cover if these APIs work correctly anyway.

test/spec/ModelerSpec.js
```diff
@@ -37,50 +37,66 @@ describe('Modeler', function() {
...
+    });
   });
 
 
   it('should not import empty definitions', function(done) {
     var xml = require('../fixtures/bpmn/empty-definitions.bpmn');
 
     // given
-    createModeler(xml, function(err, warnings, modeler) {
+    createModeler(xml).then(function(result) {
```
nikku, line 94: Can transform these tests into `awaitable` ones: Return a promise from a test and Mocha will wait for it to resolve or reject:

```javascript
    return createModeler(xml).then(function(result) {

      var modeler = result.modeler;

      // when
      return modeler.importXML(xml).catch(function(err) {

        // then
        expect(err.message).to.equal('no diagram to display');
      });
    });
```

nikku: This way we get rid of `done` entirely where we can.

test/spec/draw/TextRendererSpec.js
```diff
@@ -111,4 +111,4 @@ describe('draw - TextRenderer', function() {
 
   });
 
-});
```
nikku, line 114: Nothing changed, please undo.

test/spec/features/label-editing/TouchIntegrationSpec.js
```diff
@@ -15,21 +15,30 @@ describe('direct editing - touch integration', function() {
   function createModeler(xml, done) {
     var modeler = new Modeler({ container: container });
 
-    modeler.importXML(xml, function(err) {
-      done(err, modeler);
+    return new Promise(function(resolve) {
+
+      modeler.importXML(xml).then(function(result) {
```
nikku, line 20: Turn into chained promise.

test/spec/features/label-editing/TouchIntegrationSpec.js
```diff
@@ -15,21 +15,30 @@ describe('direct editing - touch integration', function() {
...
+      done(result.error, result.warnings);
+    });
   });
 
 
   it('should edit labels via double tap (manual test)', function(done) {
     var xml = require('./LabelEditing.bpmn');
-    createModeler(xml, done);
+    createModeler(xml).then(function(result) {
+      done(result.error, result.warnings);
```
nikku, line 40: What is the semantics of passing `warnings` as a second argument?

Better from my point of view: Get rid of `done`.

test/spec/features/modeling/LabelBoundsSpec.js
```diff
@@ -12,11 +12,15 @@ var DELTA = 2;
 
 describe('label bounds', function() {
 
-  function createModeler(xml, done) {
+  function createModeler(xml) {
     var modeler = new Modeler({ container: container });
 
-    modeler.importXML(xml, function(err, warnings) {
-      done(err, warnings, modeler);
+    return new Promise(function(resolve) {
```
nikku, line 18: Chain promises.

test/spec/features/modeling/LabelBoundsSpec.js
```diff
@@ -247,20 +256,24 @@ describe('label bounds', function() {
...
         directEditing.complete();
 
-        modeler.saveXML({ format: true }, function(err, result) {
+        modeler.saveXML({ format: true }).then(function(result) {
+
+          var xml = result.xml;
 
           // strip spaces and line breaks after '>'
-          result = result.replace(/>\s+/g,'>');
+          xml = xml.replace(/>\s+/g,'>');
```
nikku, line 264: :+1:

test/spec/features/modeling/LabelBoundsSpec.js
```diff
@@ -271,7 +284,10 @@ describe('label bounds', function() {
 
       var xml = require('./LabelBoundsSpec.simple.bpmn');
 
-      createModeler(xml, function(err, warnings, modeler) {
+      createModeler(xml).then(function(result) {
```
nikku, line 287: Chain Promises.

test/spec/features/modeling/LabelBoundsSpec.js
```diff
@@ -313,20 +333,23 @@ describe('label bounds', function() {
       // strip windows line breaks (if any)
       xml = xml.replace(/\r/g, '');
 
-      createModeler(xml, function(err, warnings, modeler) {
+      createModeler(xml).then(function(result) {
```
nikku, line 336: Chain :chains:

test/spec/features/ordering/BpmnDiOrderingSpec.js
```diff
@@ -185,10 +185,9 @@ describe('features/modeling - di ordering', function() {
 
 // helper
 function expectDiOrder(expectedOrder, done) {
-  getBpmnJS().saveXML({ format: true }, function(error, xml) {
-    if (error) {
-      return done(error);
-    }
+  getBpmnJS().saveXML({ format: true }).then(function(result) {
```
nikku, line 188: Make function awaitable (return promise) and :chains:

lib/import/Importer.js
```diff
@@ -60,22 +69,26 @@ export function importBpmnDiagram(diagram, definitions, bpmnDiagram, done) {
...
-    });
-  } catch (e) {
-    error = e;
-  }
+      eventBus.fire('import.render.complete', {
+        error: error,
+        warnings: warnings
+      });
+
+      resolve({ warnings: warnings });
```
nikku, line 87: To match `reject` and as a general good practice always `return resolve(...)`.

test/spec/import/ImporterSpec.js
```diff
@@ -41,24 +40,29 @@ describe('import - Importer', function() {
...
-    if (isFunction(diagramId)) {
-      done = diagramId;
-      diagramId = null;
-    }
+  function runImport(diagram, xml, diagramId) {
 
     var moddle = new BpmnModdle();
 
-    moddle.fromXML(xml).then(function(result) {
+    return new Promise(function(resolve, reject) {
```
nikku, line 47: CHAIN :chains:

test/spec/import/ImporterSpec.js
```diff
@@ -89,11 +93,14 @@ describe('import - Importer', function() {
       });
 
       // when
-      runImport(diagram, xml, function(err, warnings) {
+      runImport(diagram, xml).then(function() {
```
nikku, line 96: Simply return `runImport(...).then(...)` instead. Mocha handles the rest. Get rid of `done()`.

oguzeroglu:
Compatibility check is performed under CompatibilityUtil.js -> https://github.com/bpmn-io/bpmn-js/pull/1311/commits/669b5c9b84140a7311a602ff6499fbdb7186a8e9

nikku:
> Did not find the place where we check for legacy callback support.

What I meant is: 

* Where do we test that the legacy callback support still works? Where are the new test cases for that?

lib/import/Importer.js
barmac:
```suggection
      return resolve({ warnings: warnings });
```

To make it easier to fix :)

test/helper/index.js
```diff
@@ -118,7 +118,13 @@ export function bootstrapBpmnJS(BpmnJS, diagram, options, locals) {
...
     setBpmnJS(instance);
 
-    instance.importXML(diagram, done);
+    return new Promise(function(resolve) {
+      instance.importXML(diagram).then(function(result) {
+        resolve({ error: null, warnings: result.warnings });
+      }).catch(function(err) {
+        resolve({ error: err, warnings: err.warnings });
+      });
+    });
```
barmac, line 127: How about if we do chaining here?

```suggestion
    return instance.importXML(diagram)
      .then(function(result) {
        return({ error: null, warnings: result.warnings });
      })
      .catch(function(err) {
        resolve({ error: err, warnings: err.warnings });
      });
```

test/integration/SimpleModelingSpec.js
```diff
@@ -18,7 +18,10 @@ describe('scenario - simple modeling', function() {
     var modeler = new Modeler({ container: container });
 
     // when
-    modeler.createDiagram(function(err) {
+    modeler.createDiagram().then(function() {
+
+      done();
+    }).catch(function(err) {
 
       done(err);
```
barmac, line 26:
```suggestion
return modeler.createDiagram();
```

Mocha can handle promises. You also need to remove `done` here.

test/integration/model/BpmnModdleSpec.js
```diff
@@ -22,11 +22,9 @@ describe('bpmn-moddle', function() {
         '</bpmn2:definitions>';
 
       // when
-      parse(xml, function(err, definitions) {
+      parse(xml).then(function(result) {
```
barmac, line 25: Again, we can return the whole promise and if the `expect`ations are not met, promise rejection will cause the test to fail.

barmac:
I'd love to see test cases which verify that we can still use the APIs old way.

oguzeroglu:
ITT: I am learning about Promise chaining 👶

lib/BaseViewer.js
```diff
@@ -65,6 +68,21 @@ export default function BaseViewer(options) {
 
 inherits(BaseViewer, Diagram);
 
+/**
+* The importXML result.
+*
+* @typedef {Object} ImportXMLResult
+*
+* @property {Array<String>} warnings
```
barmac, line 76:
```javascript
* @property {Array<string>} warnings
```
Let's make it lowercase to not break recent @philippfromme work (https://github.com/bpmn-io/bpmn-js/commit/7a67ff95b2c06669d449f2751233e2aec6c7c61e)

lib/BaseViewer.js
```diff
@@ -65,6 +68,21 @@ export default function BaseViewer(options) {
...
+*/
+
+/**
+* The importXML error.
+*
+* @typedef {Error} ImportXMLError
+*
+* @property {Array<String>} warnings
```
barmac, line 84:
```suggestion
* @property {Array<string>} warnings
```
Same as above.

lib/BaseViewer.js
```diff
@@ -87,52 +105,93 @@ inherits(BaseViewer, Diagram);
...
   });
-};
+});
+
+/**
+* The importDefinitions result.
+*
+* @typedef {Object} ImportDefinitionsResult
+*
+* @property {Array<String>} warnings
```
barmac, line 185: Same as in the first file, it should be `Array<string>` as we are usin array of primitive values.

lib/BaseViewer.js
```diff
@@ -196,32 +289,64 @@ BaseViewer.prototype.open = function(bpmnDiagramOrId, done) {
...
+    });
+  });
+});
+
+/**
+ * The saveXML result.
+ *
+ * @typedef {Object} SaveXMLResult
+ *
+ * @property {String} xml
```
barmac, line 348:
```suggestion
 * @property {string} xml
```

nikku:
Regarding https://github.com/bpmn-io/bpmn-js/pull/1311/commits/e216c6c09f63387b995edb340c0341b919f681e2: Did you re-create these tests or did you take them from `bpmn-js@6.4.2`?

I was not sure as I saw some test using a mixture of promises and `done`.

In the best case, we simply copy the `v6.4.2` tests, moved to a section `callback style (deprecated)` and they should still work, right?

oguzeroglu:
Yes they should work. Should we proceed like that?

Do we want to copy all the existing tests in the develop branch?

barmac:
There a couple of more places where we use wrapped `String` in JSDoc. Let's fix that before we merge this.

oguzeroglu:
Thanks for the feedback @barmac, I fixed the jsdoc String types.

barmac:
OK, so IMO the last thing that needs to be fixed is missing `BREAKING CHANGES` section in the commit messages. I'd also consider squashing the `promisify` commits as they effectively are part of the same change with the same breaking change. It would be also nice to have jsdoc fixes incorporated as the fix applies to changes within this PR.

oguzeroglu:
@barmac Done! Please check 304213b46ff82641b3f642f5a9b44a35184d4be6

@nikku Any other comments? IMO this PR is good to go now!

barmac:
Looks good to me.

nikku:
Looks good. Do not merge yet. We'll cut a minor feature release first in the scope of https://github.com/camunda/camunda-modeler/issues/1776.

nikku:
(Happening today _or_ tomorrow).

lib/util/CompatibilityUtil.js
```diff
@@ -0,0 +1,24 @@
...
+
+// Wraps APIs to check:
+// 1) If a callback is passed -> Warn users about callback deprecation.
+// 2) If Promise class is implemented in current environment.
+export function wrapForCompatibility(api) {
+
+  return function() {
+
+    if (!window.Promise) {
+      throw new Error('Promises is not supported in this environment. Consider polyfilling.');
```
nikku, line 11: Consider polyfilling => Please polyfill them. Maybe?

oguzeroglu:
Callbacks are completely moved into compatibility layer. Also all the legacy tests are move under one single "describe" scope.

lib/util/CompatibilityUtil.js
```diff
@@ -0,0 +1,46 @@
...
+      // methods like pop() and slice() do not exist within arguments.
+      delete arguments[argLen - 1];
+
+      api.apply(this, arguments).then(function(result) {
+
+        var firstKey = Object.keys(result)[0];
+
+        // The APIs we are wrapping all resolve a single item depending on the API.
+        // For instance, importXML resolves { warnings } and saveXML returns { xml }.
+        // That's why we can call the callback with the first item of result.
```
barmac, line 35: Why do we actually wrap the result in an object? Maybe I missed sth...

lib/util/CompatibilityUtil.js
```diff
@@ -0,0 +1,46 @@
...
+        return callback(null, result[firstKey]);
+      }).catch(function(err) {
```
barmac, line 37: Side effect of the `catch` block is that we also catch errors within the `callback`. I'd suggest to use the second callback in `then`, i.e.
```javascript
promise.then(result => { callback(null, result); }, error => { callback(error); });
```

lib/util/CompatibilityUtil.js
```diff
@@ -0,0 +1,46 @@
...
+        'Passing callbacks to ' + api.name + ' is deprecated and will be removed in a future major release.' +
+        'Please switch to promises, cf. https://bpmn.io/l/moving-to-promises.html'
+      );
+
+      // We want to call the API without the callback.
+      // Arguments is not an Array, but an array like object. That's why
+      // methods like pop() and slice() do not exist within arguments.
+      delete arguments[argLen - 1];
+
+      api.apply(this, arguments).then(function(result) {
```
barmac, line 29: How about if we just use a slice of arguments? `delete` preserves the length of the array and set the item to `undefined`:
image

C.f. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Deleting_array_elements

lib/util/CompatibilityUtil.js
philippfromme: Shouldn't `importXML` resolve with `err` and `definitions`?

barmac: Currently, `done` is called with `err, warnings`
https://github.com/bpmn-io/bpmn-js/blob/develop/lib/BaseViewer.js#L89

lib/util/CompatibilityUtil.js
barmac: So then it would be:
```suggestion
      api.apply(this, Array.prototype.slice.call(arguments, 0, -1)).then(function(result) {
```

lib/util/CompatibilityUtil.js
oguzeroglu: For definitions we have importDefinitions.

We wrap the result in an object as a design choice. We figured it'd be easier like that to add stuff later.

philippfromme: Ah, I forgot.

barmac: OK

lib/util/CompatibilityUtil.js
barmac: or even better:
```suggestion
      var argsWithoutCallback = Array.prototype.slice.call(arguments, 0, -1);

      api.apply(this, argsWithoutCallback).then(function(result) {
```

oguzeroglu:
Thanks for the suggestions @barmac, I like them so squished them into the commit :)

nikku:
What is left here to complete this one?

oguzeroglu:
Someone who presses on the "Approve" button 🍰

oguzeroglu:
Ah it turns out that there's a conflict. Let me fix it.

lib/BaseViewer.js
```diff
@@ -241,47 +327,62 @@ BaseViewer.prototype.open = function(bpmnDiagramOrId, done) {
...
+          error: null,
+          xml: xml
+        }) || xml;
+
+        self._emit('saveXML.done', {
+          error: null,
+          xml: xml
+        });
+      } catch (e) {
+        console.error('error in saveXML life-cycle listener', e);
```
barmac, line 368: So I'd suggest that we reject here.

barmac: Otherwise the API user will have to do a null check besides error handling.

barmac: BTW the `saveXML.done` might never fire, but that's a different story.

nikku:
I've found one more thing, too:

In the past, we've passed `warnings` via `context` to `import.done`, i.e. we emitted an event `{ error, context = { warnings, ... }` (cf. https://github.com/bpmn-io/bpmn-js/blob/develop/lib/BaseViewer.js#L114).

Our API upstream changed in a way that we pass `error` and `warnings` directly in the parse result. Should we adjust the API accordingly, i.e. expose error and warnings on the same level: `{ error, warnings, ...and other things in context }`? We could keep the context + warnings for compatibility purposes but warn whenever it is being accessed. What do you think?

lib/BaseViewer.js
oguzeroglu: I disagree. The original API does not call done with err here. This PR should only promisify APIs but should not change the existing state.

barmac: Let's wait until somebody stumbles upon it then ¯\_(ツ)_/¯

nikku: @barmac If you provide a failing test case we can fix that particular issue in another PR.

barmac: All right, I will do that.

barmac:
Do you suggest that we change the event payload for `import.parse.complete` or what? Add the context to `import.done`?

nikku:
Yes. I suggest that we emit `import.parse.complete` with `{ error, warnings, elementsById, references, ...}` rather than `{ error, context: { warnings, elemetsById } }`.

barmac:
I'm ok with that. Let's just make sure that once we change it the breaking change is documented.

nikku:
I'll look into the remainder of this issue on Monday.

oguzeroglu:
@nikku @barmac I modified the `import.parse.complete` behavior with this commit: ec610310b6d92ffa4db2ec72a5bda3007d1146ef

oguzeroglu:
Conflicts fixed

oguzeroglu:
Any updates? IMO lets merge it 🎃

barmac:
@philippfromme is in the middle of release as I can see, but other than that +1

oguzeroglu:
🐵

Issue #1356, opened 2020-10-08
How to get all translations for bpmn-js and  bpmn-js-properties-panel?


gotounix:
I found `translation-reporter.js` in `test/config` folder, but I don't know how to use it.
Is there some docs telling how to get all string fields for translating in `bpmn-js` and `bpmn-js-properties-panel`?
Thanks.

nikku:
What would you like to achieve?

Up to date translations can be found [here](https://github.com/bpmn-io/bpmn-js-i18n).

nikku:
I kindly ask you to use our issue template to file your issues. For simple questions [our forum](https://forum.bpmn.io/) is the right place to discuss.

gotounix:
> What would you like to achieve?
> 
> Up to date translations can be found [here](https://github.com/bpmn-io/bpmn-js-i18n).

I found the translations in [here](https://github.com/bpmn-io/bpmn-js-i18n) is not accurate and complete. Its coverage is less than 80%.
And I can't found any translation template for bpmn-js-properties-panel.

pinussilvestrus:
This [community maintained plugin](https://github.com/FlowSquad/camunda-modeler-i18n-plugin/tree/master/client/i18n-extension/languages) also offers some translation lists.

nikku:
Closing this as something along https://github.com/bpmn-io/bpmn-js/issues/1356#issuecomment-707820890 serves as a fine workaround. For further questions, please use our forum.

nikku:
As a late update to this one: https://github.com/bpmn-io/bpmn-js/pull/1364 restores the translation reporter that collects translations during a test run. Technically you can use the reporter to collect _all_ translations, i.e. modeler + properties panel, too.

As you indicated, some of our existing translations are outdated. For an up-to-date list of translatable strings, checkout https://github.com/bpmn-io/bpmn-js/pull/1364/files#diff-42ab64542245ba48c4b9b5fc3460c742f6ae7d58e7c279282a3a0d3e0e7716ed.

PR #1364, opened 2020-11-03
Resurrect translation reporter

nikku:
Related to https://github.com/bpmn-io/bpmn-js/issues/1356

---

This restores a translation reporter that we had in good old grunt times to collect translatable strings. The PR also makes sure we cover a few more existing functions, added a bunch of new tests accordingly.


docs/translations.json
```diff
@@ -41,8 +59,8 @@
   "Escalation Start Event",
   "Escalation Start Event (non-interrupting)",
   "Event Sub Process",
+  "Event based Gateway",
   "Exclusive Gateway",
-  "HELLO {you}!",
```
barmac, line 45: 💕

test/config/karma.unit.js
```diff
@@ -86,5 +89,16 @@ module.exports = function(karma) {
       },
       devtool: 'eval-source-map'
     }
-  });
+  };
+
+  if (collectTranslations) {
+    config.plugins = [].concat(config.plugins || [], 'karma-*', require('./translation-reporter'));
+    config.reporters = [].concat(config.reporters || [], 'translation-reporter');
+
```
MaxTru, line 97: Why the linebreak?

MaxTru (approved):
:muscle: 
* Looks very good 
* I verified then when running locally I generate the exact same translation file
* Nice examples for the "refactorings on the way" topic you mentioned today

PR #1373, opened 2020-11-17
Build and test distro after version bump

### LINKED ISSUE #1369: NPM - Package version not updated
__Describe the Bug__

The version of the npm package in the dist folder is not updated.
The latest version is 7.4.1, however the version that is shown in the deployed dist files is set to 7.4.0.

__Steps to Reproduce__

1. yarn add bpmn-js
2. Check files in node_modules/dist/...

=> bpmn-modeler.development.js

image

image

barmac:
Closes #1369

__Definition of Done__

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "test": "karma start test/config/karma.unit.js",
     "distro": "node tasks/build-distro.js",
     "test:distro": "node tasks/test-distro.js",
-    "preversion": "run-s distro test:distro"
+    "postversion": "run-s distro test:distro"
   },
   "repository": {
     "type": "git",


package.json
```diff
@@ -9,7 +9,7 @@
     "test": "karma start test/config/karma.unit.js",
     "distro": "node tasks/build-distro.js",
     "test:distro": "node tasks/test-distro.js",
-    "preversion": "run-s distro test:distro"
+    "prepublishOnly": "run-s distro test:distro"
```
nikku, line 12: The reason for the inverse change (https://github.com/bpmn-io/bpmn-js/commit/bd0e86d3136d14082636f6b1af7c759471d6351c) was that publishing failed with `prepublishOnly`: The OTP publish token would always timeout before the bundle completed. 

Maybe we can hook into another step _or_ simplify the build step to not include the additional tests.

barmac: Thanks for the insight. We could then use `postversion` instead. It will still be run on publish but before `npm publish` step.

nikku: Works for me, if it works for `np` + OTP, too. :+1:

barmac: It should be as `npm publish` is a separate step and only then one is asked for OTP.

nikku:
Let's try it :runner:

PR #1374, opened 2020-11-18
chore(CONTRIBUTING): document unit test formal structuring guidance

MaxTru:
This PR aims at providing some documentation which can be used as a reference when writing tests in all our libararies.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -11,6 +11,8 @@ Great to see you! Help us out by [filing bugs or feature requests](#working-with
 * [Contributing Improvements](#contributing-improvements)
     * [Setting up the Project locally](#setting-up-the-project-locally)
     * [Discussing Code Changes](#discussing-code-changes)
+    * [Adhering to the Code Style](#adhering-to-the-code-style)
+    * [Adhering to the Unit Test Style](#adhering-to-the-unit-test-style)
     * [Creating a Pull Request](#creating-a-pull-request)
 
 
@@ -58,7 +60,7 @@ For details consult our in depth [setup guide](https://github.com/bpmn-io/bpmn-j
 Create a [pull request](#creating-a-pull-request) if you would like to have an in-depth discussion about some piece of code.
 
 
-### Code Style
+### Adhering to the Code Style
 
 In addition to our automatically enforced [lint rules](https://github.com/bpmn-io/eslint-plugin-bpmn-io), please adhere to the following conventions:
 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -67,6 +67,37 @@ In addition to our automatically enforced [lint rules](https://github.com/bpmn-i
 
 __Rationale:__ People should be able to consume parts of the library with an ES module aware bundler such as [Webpack](https://webpack.js.org/) or [Rollup](https://rollupjs.org) without the need to use a transpiler such as [Babel](https://babeljs.io/).
 
+### Unit Tests
+
+In order to retrieve a sign-off for your contribution, it needs to be sufficiently and well tested. Please structure your unit tests into [**given**, **when** and **then**](https://gist.github.com/nikku/1dc2edb553565238c7bf). To increase overall readability and understandability please also leave two empty lines before each `describe(...)` and `it(...)` block unless it is the first statement in the current indentation level. Also see this example:
+
+```
```
nikku, line 74: Rather than quoting long lines of tests, maybe it makes sense to link into the existing code base (which is fairly well tested in a consistent manner). 

If we wanted to explain testing in detail here (which we don't, I guess) we'd need to cover so many more things.

.github/CONTRIBUTING.md
```diff
@@ -67,6 +67,37 @@ In addition to our automatically enforced [lint rules](https://github.com/bpmn-i
 
 __Rationale:__ People should be able to consume parts of the library with an ES module aware bundler such as [Webpack](https://webpack.js.org/) or [Rollup](https://rollupjs.org) without the need to use a transpiler such as [Babel](https://babeljs.io/).
 
+### Unit Tests
```
nikku, line 70: We could link the test section here: https://github.com/bpmn-io/bpmn-js/pull/1374/files#diff-98e64bc1cd2db9333c6effe87bbe0d6dfe8714aba4c6bde45aa037fe0796e44cR112.

.github/CONTRIBUTING.md
```diff
@@ -67,6 +67,37 @@ In addition to our automatically enforced [lint rules](https://github.com/bpmn-i
...
+
+In order to retrieve a sign-off for your contribution, it needs to be sufficiently and well tested. Please structure your unit tests into [**given**, **when** and **then**](https://gist.github.com/nikku/1dc2edb553565238c7bf). To increase overall readability and understandability please also leave two empty lines before each `describe(...)` and `it(...)` block unless it is the first statement in the current indentation level. Also see this example:
```
nikku, line 72: > leave two empty lines before each `describe(...)` and `it(...)` block

Maybe "separate `it` and `describe` blocks on the same level with two empty lines"?

nikku (changes requested):
Left some comments.

I like adding that section, would love to have it shorter and simpler though. More text is less likely to be read.

.github/CONTRIBUTING.md
MaxTru: But that would not include the two lines prior to a `it` or `describe` block after another `setup` block, would it?

.github/CONTRIBUTING.md
MaxTru: Good idea, will work on it

.github/CONTRIBUTING.md
nikku: "separate `before`, `it` and `describe` blocks on the same level with two empty lines" would work for me just fine.

Generally speaking I'd be fine with some ambiguity here. If we really would like to enforce it, we could [create an eslint rule](https://github.com/bpmn-io/eslint-plugin-bpmn-io) that fixes things auto-magically.

nikku: Fine to keep it as is, too.

What would be more important for me is that users see actual well written test cases linked => if I'm getting into a project I should contribute in the spirit of the project, even if no special style guide exists.

.github/CONTRIBUTING.md
MaxTru: The link you provided just links to my commit? I am not sure I get what you mean.

.github/CONTRIBUTING.md
MaxTru: Added examples :+1:

.github/CONTRIBUTING.md
nikku: Points to last section: 

```
...
* The code is tested
...
```

MaxTru: :+1: Done

nikku (approved):
Works for me.

General assessment: We could spend an hour or two to rewrite this document entirely.

PR #1386, opened 2020-12-07
feat(replace-menu): update label of collapsed pool

pinussilvestrus:

* Indicate the action in a clearer, more standard-compliant way --> replaces an expanded pool with an empty "black box" + removing its content.
* We will not update the action identifier (`replace-with-collapsed-pool`) to avoid breaking changes.
* Make it possible to retrieve label via function

Related to camunda/camunda-modeler#2022

__Acceptance Criteria__

* [ ] Corresponds to the concept 
* [ ] Corresponds to the design 

__Definition of Done__

### DIFF
--- a/lib/features/popup-menu/ReplaceMenuProvider.js
+++ b/lib/features/popup-menu/ReplaceMenuProvider.js
@@ -390,10 +390,15 @@ ReplaceMenuProvider.prototype._createMenuEntry = function(definition, element, a
     return replaceElement(element, definition.target);
   };
 
+  var label = definition.label;
+  if (label && typeof label === 'function') {
+    label = label(element);
+  }
+
   action = action || replaceAction;
 
   var menuEntry = {
-    label: translate(definition.label),
+    label: translate(label),
     className: definition.className,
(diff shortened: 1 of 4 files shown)


barmac:
What about the "Expanded Pool"? Do we want to keep it that way or maybe come up with something which aligns better with "Empty Pool"?

pinussilvestrus:
For me, it would be okay to stay with "Expanded Pool", it explains clear enough what will happen. An alternative would be maybe "Filled Pool" or "Create Process inside Pool".

barmac:
How about a "Pool" without adjectives? If not that, then I would leave "Expanded Pool" as it is.

I just checked how Signavio solves that problem:

image

pinussilvestrus:
Yeah, I already thought about the Signavio way of doing it. I think it's confusing if you have an Empty Pool and in the replace menu you just found the option "Pool". Because, technically, you already on a pool, but it's just empty. An adjective would be better here, "Expanded" might be good enough then.

barmac (approved):
🚀

PR #1393, opened 2020-12-09
Add multiplicity marker to replace menu for participants

### LINKED ISSUE #533: I can model a pool with MI marker
In BPMN it is possible to mark a pool as multi instance (at least the marker looks like the parallel multi instance marker)

Should work for both, collapsed and expanded pools

azeghers:

Closes #533

__Acceptance Criteria__

* [ ] Corresponds to the concept 
* [ ] Corresponds to the design 

__Definition of Done__

### DIFF
--- a/lib/features/popup-menu/ReplaceMenuProvider.js
+++ b/lib/features/popup-menu/ReplaceMenuProvider.js
@@ -24,9 +24,10 @@ import * as replaceOptions from '../replace/ReplaceOptions';
  * This module is an element agnostic replace menu provider for the popup menu.
  */
 export default function ReplaceMenuProvider(
-    popupMenu, modeling, moddle,
+    bpmnFactory, popupMenu, modeling, moddle,
     bpmnReplace, rules, translate) {
 
+  this._bpmnFactory = bpmnFactory;
   this._popupMenu = popupMenu;
   this._modeling = modeling;
   this._moddle = moddle;
@@ -38,6 +39,7 @@ export default function ReplaceMenuProvider(
 }
 
 ReplaceMenuProvider.$inject = [
+  'bpmnFactory',
   'popupMenu',
   'modeling',
   'moddle',
@@ -268,6 +270,10 @@ ReplaceMenuProvider.prototype.getHeaderEntries = function(element) {
     headerEntries = headerEntries.concat(this._getDataObjectIsCollection(element));
   }
 
+  if (is(element, 'bpmn:Participant')) {
+    headerEntries = headerEntries.concat(this._getParticipantMultiplicity(element));
+  }
+
   if (is(element, 'bpmn:SubProcess') &&
       !is(element, 'bpmn:Transaction') &&
       !isEventSubProcess(element)) {
@@ -486,7 +492,7 @@ ReplaceMenuProvider.prototype._getLoopEntries = function(element) {
 };
 
 /**
- * Get a list of menu items containing buttons for multi instance markers
+ * Get a list of menu items containing a button for the collection marker
  *
  * @param  {djs.model.Base} element
  *
@@ -517,6 +523,46 @@ ReplaceMenuProvider.prototype._getDataObjectIsCollection = function(element) {
     }
   ];
   return dataObjectEntries;
+};
+
+/**
+ * Get a list of menu items containing a button for the participant multiplicity marker
+ *
+ * @param  {djs.model.Base} element
+ *
+ * @return {Array<Object>} a list of menu items
+ */
+ReplaceMenuProvider.prototype._getParticipantMultiplicity = function(element) {
+
+  var self = this;
+  var bpmnFactory = this._bpmnFactory;
+  var translate = this._translate;
+
+  function toggleParticipantMultiplicity(event, entry) {
+    var isActive = entry.active;
+    var participantMultiplicity;
+
+    if (!isActive) {
+      participantMultiplicity = bpmnFactory.create('bpmn:ParticipantMultiplicity');
+    }
+
+    self._modeling.updateProperties(
+      element,
+      { participantMultiplicity: participantMultiplicity });
+  }
+
+  var participantMultiplicity = element.businessObject.participantMultiplicity;
+
+  var participantEntries = [
+    {
+      id: 'toggle-participant-multiplicity',
+      className: 'bpmn-icon-parallel-mi-marker',
+      title: translate('Participant Multiplicity'),
+      active: !!participantMultiplicity,
+      action: toggleParticipantMultiplicity,
+    }
+  ];
+  return participantEntries;
 };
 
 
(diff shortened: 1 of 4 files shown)


test/spec/features/popup-menu/ReplaceMenuProviderSpec.js
```diff
@@ -188,6 +189,104 @@ describe('features/popup-menu - replace menu provider', function() {
...
+
+
+  describe('participants - multiplicity marker', function() {
+
+    beforeEach(bootstrapModeler(diagramXMLParticipants, { modules: testModules }));
+
+
+    it('should toggle on', inject(function(elementRegistry) {
+
+      // given
```
MaxTru, line 203: The 
```
      // given
      var participant = elementRegistry.get('Participant_1');

      openPopup(participant);
```

is always the same for all 4 test cases isn't it? I suggest to therefore make it part of an "forEach" block. (also applies to the old tests above probably)

MaxTru (commented):
Good job.
I tested in locally, runs well.

Only a minor comment.

lib/features/popup-menu/ReplaceMenuProvider.js
```diff
@@ -519,6 +523,38 @@ ReplaceMenuProvider.prototype._getDataObjectIsCollection = function(element) {
...
+ */
+ReplaceMenuProvider.prototype._getParticipantMultiplicity = function(element) {
+
+  var self = this;
+  var translate = this._translate;
+
+  function toggleParticipantMultiplicity(event, entry) {
+    self._modeling.updateProperties(
+      element,
+      { participantMultiplicity: !entry.active });
```
pinussilvestrus, line 541: I think we have to be careful with this. [According to the spec, this is **not** a boolean](https://github.com/bpmn-io/bpmn-moddle/blob/master/resources/bpmn/json/bpmn.json#L1970), but an element with type `bpmn: ParticipantMultiplicity`. If we store it is as a boolean, we will get in trouble while serializing (`modeler#saveXML`). 

E.g. when using the inside Camunda Modeler, I got the following error when switching to the XML Editor.

```text
/Users/niklas.kiefer/Desktop/diagram_1.bpmn] Cannot read property 'isGeneric' of undefined
    at ElementSerializer.build (webpack-internal:///../../bpmn-js/node_modules/moddle-xml/dist/index.esm.js:1226:37)
    at eval (webpack-internal:///../../bpmn-js/node_modules/moddle-xml/dist/index.esm.js:1472:30)
    at forEach (webpack-internal:///../../bpmn-js/node_modules/min-dash/dist/index.esm.js:181:16)
    at eval (webpack-internal:///../../bpmn-js/node_modules/moddle-xml/dist/index.esm.js:1460:63)
    at forEach (webpack-internal:///../../bpmn-js/node_modules/min-dash/dist/index.esm.js:181:16)
    at ElementSerializer.parseContainments (webpack-internal:///../../bpmn-js/node_modules/moddle-xml/dist/index.esm.js:1432:59)
    at ElementSerializer.build (webpack-internal:///../../bpmn-js/node_modules/moddle-xml/dist/index.esm.js:1247:10)
    at eval (webpack-internal:///../../bpmn-js/node_modules/moddle-xml/dist/index.esm.js:1472:30)
    at forEach (webpack-internal:///../../bpmn-js/node_modules/min-dash/dist/index.esm.js:181:16)
    at eval (webpack-internal:///../../bpmn-js/node_modules/moddle-xml/dist/index.esm.js:1460:63) [ error ]
```

That's because `moddle` does not know what due to the Boolean value.

pinussilvestrus: I think better would be something like

```js
var isActive = entry.active;

var participantMultiplicity;

if(isActive) {
  participantMultiplicity = bpmnFactory.create('bpmn:ParticipantMultiplicity');
}

self._modeling.updateProperties(
  element,
  { participantMultiplicity: participantMultiplicity });
);

```

pinussilvestrus: [The `BpmnRenderer`](https://github.com/bpmn-io/bpmn-js/blob/5d339676f2e6fb3b095185631d76c83cdaebca95/lib/draw/BpmnRenderer.js#L1170) is also not checking for a boolean only, but for an _existing_ object.

pinussilvestrus (changes requested):

lib/features/popup-menu/ReplaceMenuProvider.js
pinussilvestrus: Also, cf. this example diagram: https://github.com/bpmn-io/bpmn-js/blob/master/test/fixtures/bpmn/draw/pools-with-collection-marker.bpmn#L4

pinussilvestrus:
Maybe we can also align the marker with the one for activities. It looks a bit less bold.

image

lib/features/popup-menu/ReplaceMenuProvider.js
azeghers: Thanks for the catch, I was a bit too quick to implement it.
I'll look into it and update my PR :+1:

test/spec/features/popup-menu/ReplaceMenuProviderSpec.js
azeghers: I recall @nikku told me a few weeks back that when it came to testing, DRY wasn't the best guideline, as a test was supposed to be understandable as a standalone process, so it made sense to repeat some of the setup for each of them.

lib/features/popup-menu/ReplaceMenuProvider.js
pinussilvestrus: 🚀 🚀

test/spec/features/popup-menu/ReplaceMenuProviderSpec.js
MaxTru: Well when the goal is understandability, I would still argue that the understandability is not reduced when using a forEach block for this. The key content of the test is to click the icon, and that would still be included in the test hence would be understood.  
Anyway, this is just a suggestion which improves it IMO. But it is also okay the way it is.

azeghers:
Marker aligned and boolean changed to element

image

nikku:
> Marker aligned and boolean changed to element

Great to see all of these on a single screen. Is there a particular reason the vertical offsets are still different? This is not a big one but drives pixel dudes like me crazy :stuck_out_tongue_winking_eye:.

azeghers:
That's how the marker was defined for participants :man_shrugging: 
Do you want me to give it a bit more distance from the bottom?

nikku:
It is fine :see_no_evil: I guess.

pinussilvestrus (commented):
Works perfectly fine for me 👏  Good job!

I Would wait for @MaxTru to give his final feedback before merging it.

pinussilvestrus (approved):

PR #1394, opened 2020-12-16
Add step in setup file to run modeler in browser

azeghers:
This PR adds a section that documents how to execute a single test that allows to run an instance of the modeler in the browser.

### DIFF
--- a/docs/project/SETUP.md
+++ b/docs/project/SETUP.md
@@ -61,6 +61,27 @@ Execute `npm install` on each of the projects to grab their dependencies.
 Execute `npm run all` on each project. Things should be fine.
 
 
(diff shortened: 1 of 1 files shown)


nikku:
One thought I have here.

What if the user could simply execute

```
SINGLE_MODELER=1 npm run dev
```

And what you've mentioned would just work, out of the box?

nikku:
Let's merge this.

If https://github.com/bpmn-io/bpmn-js/issues/1396 yields any meaningful results, we can update our steps. 

I'd event put a one-liner, if https://github.com/bpmn-io/bpmn-js/issues/1396 produces one directly into the README.

PR #1411, opened 2021-01-22
chore: create basic GitHub action CI

nikku:
This adds a GitHub actions CI pipeline to our build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,7 @@ jobs:
       before_script:
         - ./tasks/wiredeps
       script:
-        - COVERAGE=1 TEST_BROWSERS=Firefox,PhantomJS,ChromeHeadless xvfb-run npm run all
-        - npx codecov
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -9,8 +9,7 @@ jobs:
       before_script:
         - ./tasks/wiredeps
```
nikku, line 10: We do not do that on GitHub actions right now, should we?

philippfromme: Yes, we want the same behavior we previously had on Travis.

nikku: O.k. need to add this one then.

philippfromme (approved):
Had to disable Firefox and PhantonJS on Windows. PhantomJS on Mac doesn't seem to be entirely stable but okay for now.

nikku:
Thanks for approving and looking into the windows issues. I :heart: how easy these kinds of conditional pipelines are:

image

PR #1414, opened 2021-01-26
fix(modeling): if we fix hover, we gotta fix out, too

nikku:
This ensures our FixHoverBehavior (that ensures we drop onto / connect to participants, not lanes) does fix the hover element during *.out, too.

Otherwise there is no way for other behaviors to correctly remove hover indicators.

We did properly handle that during connect already, just not during shape add or move :dash:.

---

Closes https://github.com/bpmn-io/bpmn-js/issues/1413


philippfromme (approved):
👏

PR #1417, opened 2021-03-01
Always save xml done

nikku:
This ensures we can safely listen to `saveXML.done` (`{ error, xml }`) to get notified on save errors, too.

### DIFF
--- a/lib/BaseViewer.js
+++ b/lib/BaseViewer.js
@@ -358,12 +358,12 @@ BaseViewer.prototype.saveXML = wrapForCompatibility(function saveXML(options) {
 
   var definitions = this._definitions;
 
-  return new Promise(function(resolve, reject) {
+  return new Promise(function(resolve) {
 
     if (!definitions) {
-      var err = new Error('no definitions loaded');
-
-      return reject(err);
+      return resolve({
+        error: new Error('no definitions loaded')
+      });
     }
 
     // allow to fiddle around with definitions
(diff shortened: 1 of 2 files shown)


lib/BaseViewer.js
```diff
@@ -375,25 +375,27 @@ BaseViewer.prototype.saveXML = wrapForCompatibility(function saveXML(options) {
...
+
+    self._emit('saveXML.done', result);
+
+    var error = result.error;
+
+    if (error) {
+      return Promise.reject(error);
+    }
+
+    return result;
```
pinussilvestrus, line 398: that's pretty cool 👏  Nice improvement

pinussilvestrus (approved):

pinussilvestrus:
don't how about the test coverage, [those lines](https://github.com/bpmn-io/bpmn-js/pull/1417/files#diff-ef3f247320628594b1c9d7aed9e505d1bcfe87219c4adf4b8015c375603a6c7fR364-R365) (no definitions) hadn't any test case beforehand. Maybe a good chance to improve it now 🙂

nikku:
You are right, unfortunately :wink:.

Will follow up with an additional test if I find the time to do so.

nikku:
I added the additional test: https://github.com/bpmn-io/bpmn-js/pull/1417/commits/91ff537514cc22ba2e15f1122b2de60950d18217

PR #1437, opened 2021-04-21
feat(auto-place): scroll canvas to new element

marstamm:
Automatically scrolls the added element into view

recording (2)

related to https://github.com/camunda/camunda-modeler/issues/1249

### DIFF
--- a/lib/features/auto-place/BpmnAutoPlace.js
+++ b/lib/features/auto-place/BpmnAutoPlace.js
@@ -1,4 +1,4 @@
-import { getNewShapePosition } from './BpmnAutoPlaceUtil';
+import { getNewShapePosition, getScrollOffset } from './BpmnAutoPlaceUtil';
 
 
 /**
@@ -6,13 +6,21 @@ import { getNewShapePosition } from './BpmnAutoPlaceUtil';
  *
  * @param {EventBus} eventBus
  */
-export default function AutoPlace(eventBus) {
+export default function AutoPlace(eventBus, canvas) {
   eventBus.on('autoPlace', function(context) {
     var shape = context.shape,
         source = context.source;
 
     return getNewShapePosition(source, shape);
   });
+
+
+  eventBus.on('autoPlace.end', function(event) {
+    var viewBox = canvas.viewbox();
+    var shape = event.shape;
+
+    canvas.scroll(getScrollOffset(shape, viewBox));
+  });
 }
 
-AutoPlace.$inject = [ 'eventBus' ];
+AutoPlace.$inject = [ 'eventBus', 'canvas' ];
--- a/lib/features/auto-place/BpmnAutoPlaceUtil.js
+++ b/lib/features/auto-place/BpmnAutoPlaceUtil.js
@@ -136,3 +136,25 @@ export function getDataElementPosition(source, element) {
 
   return findFreePosition(source, element, position, generateGetNextPosition(nextPositionDirection));
 }
+
+/**
+ * Always have the new element in frame
+ */
+export function getScrollOffset(element, boundingBox) {
+  var elementTrbl = asTRBL(element);
+  var bbTrbl = asTRBL(boundingBox);
+
+  // Bigger padding to accommodate the context pad
+  var paddingRight = 80;
+  var padding = 10;
+
+  var dRight = Math.max(0, elementTrbl.right - bbTrbl.right + paddingRight);
+  var dLeft = Math.min(0, elementTrbl.left - bbTrbl.left - padding);
+  var dBottom = Math.max(0, elementTrbl.bottom - bbTrbl.bottom + padding);
+  var dTop = Math.min(0, elementTrbl.top - bbTrbl.top - padding);
+
+  var dx = dRight || dLeft;
+  var dy = dBottom || dTop;
+
+  return { dx: -dx, dy: -dy };
+}
--- a/test/spec/features/auto-place/BpmnAutoPlaceSpec.js
+++ b/test/spec/features/auto-place/BpmnAutoPlaceSpec.js
@@ -4,6 +4,7 @@ import {
 } from 'test/TestHelper';
 
 import autoPlaceModule from 'lib/features/auto-place';
+import { getScrollOffset } from 'lib/features/auto-place/BpmnAutoPlaceUtil';
 import coreModule from 'lib/core';
 import labelEditingModule from 'lib/features/label-editing';
 import modelingModule from 'lib/features/modeling';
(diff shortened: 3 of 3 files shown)


lib/features/auto-place/BpmnAutoPlaceUtil.js
```diff
@@ -135,4 +135,26 @@ export function getDataElementPosition(source, element) {
...
+
+/**
+ * Always have the new element in frame
+ */
+export function getScrollOffset(element, boundingBox) {
+  var elementTrbl = asTRBL(element);
+  var bbTrbl = asTRBL(boundingBox);
+
+  // Bigger padding to accommodate the context pad
+  var paddingRight = 80;
```
pinussilvestrus, line 148: where do we get this value from?

test/spec/features/auto-place/BpmnAutoPlaceSpec.js
```diff
@@ -255,6 +256,73 @@ describe('features/auto-place', function() {
...
+  describe('scroll handling', function() {
+
+    it('scrolls element into view', function() {
+
+      // given
+      var element = { x: 0, y: 0, width: 20, height: 20 };
+      var viewBox = { x: 100, y: 100, width: 100, height: 100 };
+
+      // when
+      var scrollArguments = getScrollOffset(element, viewBox);
```
pinussilvestrus, line 269: So we are testing the util, but not the actual event handler? Is this intended?

lib/features/auto-place/BpmnAutoPlace.js
```diff
@@ -1,18 +1,26 @@
...
     var shape = context.shape,
         source = context.source;
 
     return getNewShapePosition(source, shape);
   });
+
+
+  eventBus.on('autoPlace.end', function(event) {
+    var viewBox = canvas.viewbox();
+    var shape = event.shape;
```
pinussilvestrus, line 20: Just some personal preference: let's try to stick with our current style

```js
var viewBox = canvas.viewBox(),
    shape = event.shape;
```

also maps to other examples below

lib/features/auto-place/BpmnAutoPlaceUtil.js
```diff
@@ -135,4 +135,26 @@ export function getDataElementPosition(source, element) {
   };
 
   return findFreePosition(source, element, position, generateGetNextPosition(nextPositionDirection));
+}
+
+/**
+ * Always have the new element in frame
+ */
+export function getScrollOffset(element, boundingBox) {
+  var elementTrbl = asTRBL(element);
```
pinussilvestrus, line 144: cf. above, I'd stick with the current style of multi-variable declaration

lib/features/auto-place/BpmnAutoPlace.js
```diff
@@ -1,18 +1,26 @@
...
+import { getNewShapePosition, getScrollOffset } from './BpmnAutoPlaceUtil';
 
 
 /**
  * BPMN auto-place behavior.
  *
  * @param {EventBus} eventBus
  */
-export default function AutoPlace(eventBus) {
+export default function AutoPlace(eventBus, canvas) {
```
pinussilvestrus, line 9: can you maybe explain, why this is a bpmn-js but no generic diagram-js concern?

pinussilvestrus (commented):
I like the improvement that the element is now visible on the canvas, no matter what 👍 

Kapture 2021-04-21 at 11 45 10

lib/features/auto-place/BpmnAutoPlace.js
nikku: One could imagine this is a general create behavior. Elements created are always moved into sight. Or what exactly do you refer to here @pinussilvestrus?

marstamm: I see the benefits of moving this PR to diagram-js, since we also moved the AutoPlace behavior there with https://github.com/bpmn-io/dmn-js/issues/470

This would also add this feature to DMN modeling, which makes sense to me

lib/features/auto-place/BpmnAutoPlaceUtil.js
marstamm: It's the css width of the context-pad (72px) + additional padding

lib/features/auto-place/BpmnAutoPlace.js
pinussilvestrus: So we have two things here

a) is it a diagram-js concern? I'd vote for **yes**
b) is this an auto-place (only) behavior? Good question, I'd expect that this also happens on creating any kind of element, no matter when this happens (auto-place, creating from the palette, ...)

lib/features/auto-place/BpmnAutoPlaceUtil.js
pinussilvestrus: Ok got it, thanks for the clarification 👍

lib/features/auto-place/BpmnAutoPlace.js
marstamm: AFAIK, auto-place is the only way we can currently place elements outside of the viewbox. When creating elements from the pallet, the drag operation automatically scrolls the canvas already

lib/features/auto-place/BpmnAutoPlaceUtil.js
marstamm: I also discussed it with @philippfromme and we decided to make the padding consistent (80 to all sides) to cover most use-cases

lib/features/auto-place/BpmnAutoPlace.js
nikku: @marstamm We could argue that is the case, even though we do allow anyone to _model via API_: https://cdn.statically.io/gh/nikku/2021-token-simulation/v0.0.2/presentation.html#22

Used by some of our side projects, e.g. https://github.com/bpmn-io/bpmn-js-cli and https://github.com/nikku/bpmn-js-cli-modeling-dsl.

We could argue that auto place is the natural higher level API that ships with such behavior. People and robots (cf. [WASDENN](https://bpmn.io/blog/posts/2021-wasdenn-ai-modeling-assistant.html)) are free to implement this this in a similar fashion.

nikku: One idea: Can we make this "ensure visible" a thing on the `canvas` in diagram-js? Scroll to an element is a common use case people ask us about on our forums. It is also being used in [our search feature already](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/search-pad/SearchPad.js#L214). Moving it into a shared place and exposing it as an API would allow a broader user base to benefit.

marstamm: Great Idea, let's move it into the canvas :+1:

marstamm: @nikku , I'm not sure if we should replace the behavior in the search feature, as the current implementation is "center on screen" and not "scroll until visible". On the other hand, changing the behavior would closer resemble a text search, as no scrolling is happening when the element is already visible. What do you think?

nikku: > On the other hand, changing the behavior would closer resemble a text search, as no scrolling is happening when the element is already visible. 

I think it makes a lot of sense to scroll a little bit less. The search result display is already very verbose. So :+1: from my perspective to scroll until visible only.

We may also consider to add a little bit more padding around the to be visible element. It does make sense to show it with context (i.e. what is behind). Would appreciate your input / if you experiment with it a little bit.

marstamm:
closed in favor of https://github.com/bpmn-io/diagram-js/pull/545

PR #1441, opened 2021-04-30
Migrate to GitHub actions

barmac:


pinussilvestrus (commented):
Looks super clean to me 👍  Maybe we should clean up outdated scripts as well: https://github.com/bpmn-io/bpmn-js/blob/develop/tasks/wiredeps

barmac:
Thanks for notice. Now it should be all right.

.github/workflows/INTEGRATION.yml
```diff
@@ -0,0 +1,39 @@
+name: INTEGRATION
```
nikku, line 1: Any reason why we do have two separate post release builds? I would have expected INTEGRATION and POST_RELEASE to be the same thing, just one matching a less strict tag.

tasks/stages/update-demo
```diff
@@ -9,7 +9,7 @@ CLONE_DIR="$WORKDIR/bpmn-io-demo"
 # create work dir
 mkdir -p "$WORKDIR"
 
-git clone --depth=1 "https://$GITHUB_AUTH@github.com/$BPMN_IO_DEMO_ENDPOINT.git" "$CLONE_DIR"
+git clone --depth=1 "https://$GH_AUTH@github.com/$BPMN_IO_DEMO_ENDPOINT.git" "$CLONE_DIR"
```
nikku, line 12: Any reason we rename here?

barmac: GitHub prevents us from setting env variables which start with `GITHUB_`. Cf. https://docs.github.com/en/actions/reference/environment-variables#naming-conventions-for-environment-variables

.github/workflows/INTEGRATION.yml
barmac: That's because of the tags pattern. Anyway, I agree that it might be confusing. Let me try to fix that.

tasks/stages/update-demo
nikku: Thanks. Got it.

.github/workflows/INTEGRATION.yml
nikku: https://github.com/bpmn-io/bpmn-js/pull/1441/commits/4b65f512c90b6987ae3bbe92e18d1e11b611f91c solves it in a less confusing manner I believe.

nikku (approved):

PR #1444, opened 2021-05-04
chore: use `deps` as commit type for updates

barmac:


barmac:
I have no idea why there is a linting problem.

nikku (approved):

PR #1450, opened 2021-05-17
test: support Promise as return value in `inject`

barmac:
This makes sure we can create async tests. Without the `return`, a test failure would end up as an unhandled Promise rejection.

So this should work in the new setup and doesn't work currently:

```js
it('should fail', inject(async function(elementRegistry) {

  expect(elementRegistry).to.exist;

  return Promise.reject(new Error('should fail'));
}));
```

### DIFF
--- a/test/helper/index.js
+++ b/test/helper/index.js
@@ -226,7 +226,7 @@ export function inject(fn) {
       );
     }
 
-    BPMN_JS.invoke(fn);
+    return BPMN_JS.invoke(fn);
   };
 }
 
--- a/test/spec/helper/InjectSpec.js
+++ b/test/spec/helper/InjectSpec.js
@@ -1,0 +1,66 @@
+import {
+  bootstrapModeler,
+  inject
+} from 'test/TestHelper';
+
+import coreModule from 'lib/core';
+import { expect } from 'chai';
+
+
+describe('helper - inject', function() {
+
+  var diagramXML = require('../../fixtures/bpmn/simple.bpmn');
+
+  beforeEach(bootstrapModeler(diagramXML, {
+    modules: [
+      coreModule
+    ]
+  }));
+
+
+  it('should work with Promise as return value', function() {
+
+    // given
+    var expected = 'resolved';
+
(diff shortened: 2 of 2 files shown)


test/helper/index.js
```diff
@@ -226,7 +226,7 @@ export function inject(fn) {
       );
     }
 
-    BPMN_JS.invoke(fn);
+    return BPMN_JS.invoke(fn);
```
nikku, line 229: What would be great is an actual test case that verifies this capability within bpmn-js.

We could also think about adding the same thing (in terms of infrastructure) to diagram-js.

nikku (changes requested):
Test missing that verifies the capability exists.

barmac:
I've added tests via https://github.com/bpmn-io/bpmn-js/pull/1450/commits/57540c5e12c56dffbb386f1b56bb8a7fe7cb40c1

barmac:
The CI fails because PhantomJS does not support async functions. Is there any reason to keep it since we dropped IE11 support?

test/spec/helper/InjectSpec.js
```diff
@@ -0,0 +1,110 @@
...
+
+    // then
+    return test().then(function(result) {
+
+      expect(result).to.eql(expected);
+    });
+  });
+
+
+  it('should work with async function', async function() {
```
nikku, line 42: We do not need to have this test. Returning promises, tested above is equivalent to `async fn()`.

test/spec/helper/InjectSpec.js
```diff
@@ -0,0 +1,110 @@
...
+
+      return Promise.reject(expected);
+    });
+
+    // then
+    return test().then(onResolved, onRejected);
+  });
+
+
+  it('should handle async error', function() {
```
nikku, line 86: Drop test.

nikku (changes requested):
Proposal: Drop `async function() { ...` tests and we're good.

nikku (approved):

PR #1457, opened 2021-05-28
Append DataStoreReference to first available process and disallow action if there are none

### LINKED ISSUE #1456: Data Store Reference is not serialized if the first participant is an empty pool
__Describe the Bug__

Data Store Reference is not created in the XML if the first participant is an empty pool.

__Steps to Reproduce__

1. Create a participant and morph it into an empty pool
2. Place a Data Store shape on diagram
3. Save and reopen diagram.

__Expected Behavior__

Two options:

1. DataStoreReference should be a child of the first process referenced in any participant.
2. Operation should be disallowed if there are no participants with processRef.

__Environment__

 - Browser: any
 - OS: any
 - Library version: 8.6.0

barmac:
Closes #1456

### DIFF
--- a/lib/features/modeling/behavior/DataStoreBehavior.js
+++ b/lib/features/modeling/behavior/DataStoreBehavior.js
@@ -2,7 +2,10 @@ import inherits from 'inherits';
 
 import CommandInterceptor from 'diagram-js/lib/command/CommandInterceptor';
 
-import { is } from '../../../util/ModelUtil';
+import {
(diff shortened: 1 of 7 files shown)


barmac:
It seems that `wiredeps` script fails to handle this PR.

barmac:
I was able to fix wiredeps. This is ready for review.

nikku (approved):
Nice little improvement :clap:

Issue #1462, opened 2021-06-15
Plain StartEvent is created instead of Message StartEvent (non-interrupting)


0x00000001A:
__Describe the Bug__

Hi. I’m trying to add Message Start Event (non-interrupting) together with SubTask to the diagram.
The problem is the “MessageStartEvent (non-interruptig)” visually looks like “MessageStartEvent”, but actually, it’s not. I’ve compared the XML in both cases and they are the same. And after I’ve reimported the XML, everything is ok, I mean, MessageStartEvent gets the correct shape.

![On Create Drag Start](https://forum.bpmn.io/uploads/default/original/2X/6/6ea6cf3dcb72e6174ec6d64f87372df16ce07e08.png)
![On Create Drag End](https://forum.bpmn.io/uploads/default/original/2X/9/96d9e295631688b60a69b14fc792a31327f65d2a.png)

I've tried to find the way how to solve it (and you may find more info about the bug there)
(note, there was an a typo, I've specified what I've installed the 8.1.0 version, but actually it was 8.5.0)
https://forum.bpmn.io/t/add-message-start-event-non-interrupting-inside-subtask-by-default/6302

But looks like, the only one way to fix this issue - install bpmn-js@8.2.0.
In my case, I have the bpmn-js@8.5.0 installed and I've tried to upgrade to the bpmn-js@8.6.1 - bug still exists.

__Steps to Reproduce__

1. [Codesandbox with bpmn-js@8.2.0 ](https://codesandbox.io/s/event-sub-process-example-rd815)
2. [Codesandbox with bpmn-js@8.6.1](https://codesandbox.io/s/event-sub-process-example-forked-41wtv)
3. [Codesandbox with bpmn-js@8.5.0](https://codesandbox.io/s/event-sub-process-example-forked-kf00q)

__Expected Behavior__

Message StartEvent (non-interrupting) is created, instead of StartEvent.

__Environment__

 - Browser: Google Chrome
 - OS: Ubuntu 20.04
 - 8.5.0, 8.6.0, 8.6.1

barmac:
Thank you for your bug report. I was able to reproduce the issue and can confirm that the bug was introduced in v8.5.0.

I suspect that it's caused by the behaviour implemented in [this pull request](https://github.com/bpmn-io/bpmn-js/pull/1434). It seems that the `canReplace` rule for some reason decides that the message start event should be replaced.

0x00000001A:
@barmac 
[...]

PR #1463, opened 2021-06-18
Attempt to fix issue #1462 - StartEvent (non interrupting) is replaced with the plain StartEvent when element is placed on the diagram

0x00000001A:
Attempt to fix issue #1462 - StartEvent (non interrupting) is replaced with the plain StartEvent when element is placed on the diagram.

Tests (BpmnReplacePreviewSpec) are need to be fixed and/or new tests should be added.
@barmac I'll glad if you can help with test cases.

### DIFF
--- a/lib/features/modeling/behavior/ReplaceElementBehaviour.js
+++ b/lib/features/modeling/behavior/ReplaceElementBehaviour.js
@@ -4,8 +4,7 @@ import { forEach } from 'min-dash';
 
 import CommandInterceptor from 'diagram-js/lib/command/CommandInterceptor';
 
-import { isEventSubProcess } from '../../../util/DiUtil';
-
+import { is, isEventSubProcess } from '../../../util/DiUtil';
 
 /**
  * BPMN-specific replace behavior.
@@ -30,7 +29,7 @@ export default function ReplaceElementBehaviour(
         target = context.parent,
         elements = context.elements;
 
-    var canReplace = bpmnRules.canReplace(elements, target);
+    var canReplace = is(target, 'bpmn:SubProcess') && bpmnRules.canReplace(elements, target);
 
     if (canReplace) {
       this.replaceElements(elements, canReplace.replacements);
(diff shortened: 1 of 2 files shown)


barmac:
Hi,

Thanks for creating this PR. Indeed some tests in the replace preview started to fail. I'll have a look at this and perhaps the issue can be fixed.

barmac:
OK so I can see that the problem is not within the `BpmnReplacePreview` but rather in the pull request itself. The tests fail with a reason -> If you drag a non-interrupting timer start event from an event subprocess, the preview is not updated correctly. The behavior is caused strictly by the changes in `BpmnRules#canReplace`.

Please feel free to adjust the PR to fix the issue detected by the test.

0x00000001A:
@barmac 
I've double checked everything, and here what I got:
You're right, that the problem is [this pull request](https://github.com/bpmn-io/bpmn-js/pull/1434). I've tried to understand the problem that the specified PR resolves and I've ended up with the following suggestion: if this PR (that mentioned before) is about disallowing copy-paste some types of start events to the subtasks, would not be better to check if the target has "bpmn:SubTask" type?

I've just tried to add this condition (target type checking) and everything is ok, tests are passing and copy-paste typed start event is blocked as it should (according to the PR).

Actually, I've changed just one line (ReplaceElementBehaviour, line 32):
```js
var canReplace = is(target, 'bpmn:SubProcess') && bpmnRules.canReplace(elements, target);
```

What do you think about it?

barmac:
Hey, I had a look at the new solution you prepared. It turns off the ReplaceElementBehavior whenever the target is not a subprocess. This, however, makes it possible to copy a non-interrupting event to a process which is not allowed by the spec. I'd suggest to look for the solution within the rules code as it is the place where the tool decides to that the event should be replaced.

barmac:
For comparison, I created a test case which is added via https://github.com/bpmn-io/bpmn-js/pull/1469

barmac:
I am closing this for now but we can reopen the PR once you want to implement it.

PR #1464, opened 2021-06-23
Fix ordering provider error handling

nikku:
+ throw explicit exception if order could not be retrieved https://github.com/bpmn-io/bpmn-js/commit/b9dba156e028ce37e098723a613101d3a1a7096b
+ make exception handling fail-safe https://github.com/bpmn-io/bpmn-js/commit/3106e4137c1f015ad559adadb825d87b574524f9
+ various housekeeping things

----

Related [Sentry issue](https://sentry.io/organizations/camunda-modeling/issues/2436652355/?project=5218065&query=is%3Aunresolved&statsPeriod=14d)


philippfromme (commented):
What is the context? Is this about an error that would pop up somewhere else and it is now thrown in the place where it actually occurs?

nikku:
It is about errors that pop up in sentry. There seems to be rare cases where we cannot retrieve the order. 

This PR makes sure that we handle each "cannot determine order" error with a proper error message (https://github.com/bpmn-io/bpmn-js/pull/1464/commits/b9dba156e028ce37e098723a613101d3a1a7096b) _and_ ensures that we do throw the actual exception and don't fail to build it (https://github.com/bpmn-io/bpmn-js/pull/1464/commits/3106e4137c1f015ad559adadb825d87b574524f9).

philippfromme (approved):
Thanks for clarification. Nice work. 🧹

PR #1468, opened 2021-06-25
deps: update bpmn-moddle to v7.1.2

### LINKED ISSUE #1467: Cannot connect Message Flows to Call Activities
__Describe the Bug__

Message Flows where a Call Activity is either a source or a target are disallowed.

According to BPMN 2.0.2 spec, a Message Flow connects Interaction Nodes. Interaction Node is a type which includes Pools, Activities and Events:

> The InteractionNode element is used to provide a single element as the source and target Message Flow associations (see Figure 9.14, above) instead of the individual associations of the elements that can connect to Message Flows (see above). Only the Pool/Participant, Activity, and Event elements can connect to Message Flows. The InteractionNode element is also used to provide a single element for source and target of Conversation Links, see page 131.

(p. 122)

Call Activity is an Activity and therefore it should be allowed to connect it with a Message Flow given the other shape is in a different Participant.

__Steps to Reproduce__

1. Create a collaboration with call activity in one of participants
2. Try to connect the call activity to another participant.

__Expected Behavior__

Call Activity is an Activity and therefore it should be allowed to connect it with a Message Flow given the other shape is in a different Participant.

__Environment__

 - Browser: any
 - OS: any
 - Library version: any modern

barmac:
Closes #1467

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2242,9 +2242,9 @@
       "dev": true
     },
     "bpmn-moddle": {
-      "version": "7.1.1",
-      "resolved": "https://registry.npmjs.org/bpmn-moddle/-/bpmn-moddle-7.1.1.tgz",
-      "integrity": "sha512-R3xDuTklpK2NtxVDHJiecEVJvY2hKZikXdyl7oaULHMCQPfTBdid7TU/LGUgrLjA3L3ihBN9AKPgcAQ6HoPjpA==",
+      "version": "7.1.2",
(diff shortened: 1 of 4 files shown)


nikku:
We could add an integration test to verify this actually works.

barmac:
Done via https://github.com/bpmn-io/bpmn-js/pull/1468/commits/c3e2702feac71650e869101a27603a1f1ff31f70

nikku (approved):

PR #1473, opened 2021-08-05
Refactoring: decouple DI from businessObject

### LINKED ISSUE #1472: Decouple DIs from business Objects
__What should we do?__

Remove the 1:1 connection between business Objects and DIs

__Why should we do it?__

The bpmn standard allows multiple DIs (in different Diagrams) to exist for the same BPMN element. With the limitation of 1 di per BO, we prevent imports of multiple diagrams.

Therefore we should
- Remove the link of BO -> di
- Keep the link from di -> BO
- Add a link from the element/shape to the di
- Adjust Importer, Updater, etc.

---

__Breaking Changes__

The di is no longer accessible from the business Object. It can only be accessed through the element, such as `element.di` or through the utility function `getDi` in `lib/util/ModelUtil.js`.

To access DI in a safe manner from a shape or connection use the `lib/util/ModelUtil#getDi` utility.

marstamm:
closes #1472

### DIFF
--- a/lib/BaseViewer.js
+++ b/lib/BaseViewer.js
@@ -528,19 +528,6 @@ BaseViewer.prototype.clear = function() {
     return;
   }
 
-  // remove businessObject#di binding
-  //
-  // this is necessary, as we establish the bindings
-  // in the BpmnTreeWalker (and assume none are given
-  // on reimport)
-  this.get('elementRegistry').forEach(function(element) {
-    var bo = element.businessObject;
-
-    if (bo && bo.di) {
-      delete bo.di;
-    }
-  });
-
   // remove drawn elements
   Diagram.prototype.clear.call(this);
 };
--- a/lib/features/modeling/BpmnUpdater.js
+++ b/lib/features/modeling/BpmnUpdater.js
@@ -16,6 +16,7 @@ import {
 
 import {
   getBusinessObject,
+  getDi,
   is
 } from '../../util/ModelUtil';
 
@@ -307,8 +308,9 @@ BpmnUpdater.prototype.updateParent = function(element, oldParent) {
   var parentShape = element.parent;
 
   var businessObject = element.businessObject,
+      di = getDi(element),
       parentBusinessObject = parentShape && parentShape.businessObject,
-      parentDi = parentBusinessObject && parentBusinessObject.di;
+      parentDi = getDi(parentShape);
 
   if (is(element, 'bpmn:FlowNode')) {
     this.updateFlowNodeRefs(businessObject, parentBusinessObject, oldParent && oldParent.businessObject);
@@ -336,13 +338,13 @@ BpmnUpdater.prototype.updateParent = function(element, oldParent) {
     this.updateSemanticParent(businessObject.dataObjectRef, parentBusinessObject);
   }
 
-  this.updateDiParent(businessObject.di, parentDi);
+  this.updateDiParent(di, parentDi);
 };
 
 
 BpmnUpdater.prototype.updateBounds = function(shape) {
 
-  var di = shape.businessObject.di;
+  var di = getDi(shape);
 
   var target = (shape instanceof Label) ? this._getLabel(di) : di;
 
@@ -382,14 +384,17 @@ BpmnUpdater.prototype.updateFlowNodeRefs = function(businessObject, newContainme
 
 
 // update existing sourceElement and targetElement di information
-BpmnUpdater.prototype.updateDiConnection = function(di, newSource, newTarget) {
+BpmnUpdater.prototype.updateDiConnection = function(connection, newSource, newTarget) {
+  var connectionDi = getDi(connection),
+      newSourceDi = getDi(newSource),
+      newTargetDi = getDi(newTarget);
 
-  if (di.sourceElement && di.sourceElement.bpmnElement !== newSource) {
-    di.sourceElement = newSource && newSource.di;
+  if (connectionDi.sourceElement && connectionDi.sourceElement.bpmnElement !== getBusinessObject(newSource)) {
+    connectionDi.sourceElement = newSource && newSourceDi;
   }
 
-  if (di.targetElement && di.targetElement.bpmnElement !== newTarget) {
-    di.targetElement = newTarget && newTarget.di;
+  if (connectionDi.targetElement && connectionDi.targetElement.bpmnElement !== getBusinessObject(newTarget)) {
+    connectionDi.targetElement = newTarget && newTargetDi;
   }
 
 };
@@ -402,6 +407,11 @@ BpmnUpdater.prototype.updateDiParent = function(di, parentDi) {
   }
 
   if (di.$parent === parentDi) {
+    return;
+  }
+
+  // Cover the case where di.$parent === undefined and parentDi === null
+  if (!parentDi && !di.$parent) {
     return;
   }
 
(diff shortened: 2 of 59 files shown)


lib/BaseViewer.js
```diff
@@ -528,19 +528,6 @@ BaseViewer.prototype.clear = function() {
...
-  // in the BpmnTreeWalker (and assume none are given
-  // on reimport)
-  this.get('elementRegistry').forEach(function(element) {
-    var bo = element.businessObject;
-
-    if (bo && bo.di) {
-      delete bo.di;
-    }
-  });
-
```
pinussilvestrus, line 543: ❤️

marstamm:
I overlooked some issues in the create/Replace behavior of Collapsible Shapes and will fix it

lib/util/ModelUtil.js
```diff
@@ -22,4 +22,15 @@ export function is(element, type) {
  */
 export function getBusinessObject(element) {
   return (element && element.businessObject) || element;
+}
+
+/**
+ * Return the di object for a given element.
+ *
+ * @param  {djs.model.Base|ModdleElement} element
```
nikku, line 30: This should not be an instance of `ModdleElement` anymore, right? Otherwise that would be a miss-use of the API.

nikku (commented):
Great change overall.

I like in particular how straight forward this is _and_ that you introduced a utility to use consistently to retrieve the DI. :muscle:

lib/util/ModelUtil.js
marstamm: Good catch, I fixed it :heavy_check_mark:

nikku:
As this is a breaking change extensions may rely on, one last question before we merge this: What is our story for backwards compatibility / allowing users to gracefully update?

I could imagine two things:

* Clearly pharsed `BREAKING CHANGES` in the Changelog (a must!)
* Throwing a meaningful error upon accessing `businessObject.di` (optional)

marstamm:
I added the breaking changes to the changelog. 

For throwing an error: where would be a good place to start looking into it? Ideally, we would have a Accessor on the business Object. As far as I can tell, they are created in bpmn-moddle and we would have to modify our code everywhere in bpmn-js where new ones are created (import, new elements, replace(?), ...).

nikku:
We do establish the relationship [here](https://github.com/bpmn-io/bpmn-js/pull/1473/commits/4434c420c705a1b765226a70c2c891cf2b4b846a#diff-ef36d646e98da95da16cca4b5abc00040213ac30bca0ca2a809423bd9fd29328L121) and could make that a property that throws.

```
Object.definedProperty(el, 'di', {
  get: function() {
    throw 'HELLO YOU';
  }
});
```

My personal opinion is: It does make sense to be verbose about it.

marstamm:
Thanks for the hint! I decided to create the error in the `elementFactory`. This way, we also cover newly created elements and not only the imported ones.

Good thing we did it, because I also found 2 places where we still accessed the di incorrectly.

marstamm:
I hope I selected the correct files to add the tests to :crossed_fingers: 

I kept the previous commit to make it clear what has changed, the commit history should probably be cleaned up anyway before we merge it.

lib/features/modeling/BpmnUpdater.js
```diff
@@ -405,6 +410,11 @@ BpmnUpdater.prototype.updateDiParent = function(di, parentDi) {
     return;
   }
 
+  // Cover the case where di.$parent === undefined and parentDi === null
+  if (!parentDi && !di.$parent) {
+    return;
```
nikku, line 415: When does this case occur and do we have a test case for it?

nikku (commented):
I've cleaned up the commit history on this one, added additional test coverage for label wiring (was non-existing in a few places) and found two other small gotchas that I've cleaned up.

At this point, there is one last comment from my side remaining.

Looks great overall. We should proceed and get this out of the door.

lib/features/modeling/BpmnUpdater.js
marstamm: This happens when we replace a Task with an expanded sub-process. The newly created Start event has a `di.$parent = null`.

We don't test this explicitly, but tests like this check if we can replace Tasks with expanded sub-processes https://github.com/bpmn-io/bpmn-js/blob/194b963959a526e8e0805043debf43bc7fb9fdb2/test/spec/features/modeling/behavior/SubProcessStartEventBehaviorSpec.js#L29-L47

nikku: Thanks for that explaination.

But then, why did it not blow up in the past? I'll investigate and report back.

nikku: I believe [this](https://github.com/bpmn-io/bpmn-js/pull/1473/files#diff-ba4080acef0bcddd7ff56277569d664b45c9089cb8f4d949ae4548c2320c57ebR112) is the issue. We cannot blindly re-use the existing DI on replace but must continue to create a new one.

nikku: Otherwise the following will happen:

* Task (to be replaced) gets removed
* DI (of task _and_ replaced element) is being unlinked
* Replaced element does not have a proper DI reference anymore

nikku: Tests for proper DI updating after replace (`BpmnReplaceSpec`) are missing entirely and need to be added.

nikku (changes requested):
Found [the last bug](https://github.com/bpmn-io/bpmn-js/pull/1473/files/589da7e3eee0ccebc3fc9cc092c7e6c12610b8c8#r697869405) :tm:.

Please have a look. We're light on DI update test coverage, this is why this one did not blow up on your PR.

marstamm:
Found the root-cause:

When creating elements in the elementFactory, we [supported passing an Object with `di` Attributes](https://github.com/bpmn-io/bpmn-js/blob/develop/lib/features/modeling/ElementFactory.js#L94-L98). We now expect the di to be present or absent.
`attrs.di` is now used for the di moddleElement, so I broke the option to pass an Object.

We should probably fix this, but I'm not sure what's the best way:
1. Overload `attrs.di` and check if `attrs.di` is an Object or a Moddle element. Do we have utility functions to check for a Moddle element?
2. have a separate `diProperties` to be applied to the newly created `di`
3. Require the di to be created before `createBpmnElement` is called

2 and 3 would be a change to the API, but feel cleaner than overloading the attribute

nikku:
It is safe to use `ModelUtil#is` on diagram elements _and_ plain objects. I'd therefor advocate for option :one:.

nikku:
Also, we do a similar `instanceof` if not create check [in our core Modeling API](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/modeling/Modeling.js#L506) already.

test/spec/features/modeling/behavior/ReplaceElementBehaviourSpec.js
```diff
@@ -535,7 +535,7 @@ describe('features/modeling - replace element behavior', function() {
         // then
         var createdEvent = elementRegistry.get(id);
 
-        expect(createdEvent).to.eql(startEvent);
+        expect(createdEvent).to.exist;
```
marstamm, line 538: The `BpmnReplace` will now always create a new `di` and `shape`. This is the only place where we checked if it's the same, all other tests just test for an existing new shape as well:
 https://github.com/bpmn-io/bpmn-js/blob/fc2201a1bcc8d28de58ac7e477bdfba51b179a3c/test/spec/features/modeling/behavior/ReplaceElementBehaviourSpec.js#L474-L491

nikku (approved):
I've added [another minor cleanup on top](https://github.com/bpmn-io/bpmn-js/pull/1473/commits/75bc81ef6b9133b31f92ea6b68af6998ddd1fecd). 

Let us merge this and fix things that are left open later on :tada:.

PR #1476, opened 2021-08-20
ci: wait for published before executing post release steps

nikku:
Follow up to https://github.com/bpmn-io/form-js/pull/151.


barmac (approved):
👀

PR #1478, opened 2021-08-24
Update imports and export

joelknott:
Modified to use 'export default' instead of 'module.exports' so that I don't get errors stating that there is no default export.

I was receiving an error that there was no default export. It looks like most of the other handlers use 'export default' instead of 'module.exports'


nikku (approved):

nikku:
Fixed and merged via https://github.com/bpmn-io/bpmn-js/commit/8d90d4dc6391b49f9dbf6a5c8ae3ac0db8024f06. Thanks!

PR #1480, opened 2021-08-30
fix(importer): pass context during collaboration import

marstamm:

### DIFF
--- a/test/spec/import/BpmnTreeWalkerSpec.js
+++ b/test/spec/import/BpmnTreeWalkerSpec.js
@@ -5,6 +5,7 @@ import BpmnModdle from 'bpmn-moddle';
 import { find } from 'min-dash';
 
 import simpleXML from 'test/fixtures/bpmn/simple.bpmn';
+import collaboration from 'test/fixtures/bpmn/collaboration.bpmn';
 
 
 describe('import - BpmnTreeWalker', function() {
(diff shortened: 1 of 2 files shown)


test/spec/import/BpmnTreeWalkerSpec.js
```diff
@@ -138,13 +168,13 @@ function createWalker(listeners) {
...
-      listeners.element && listeners.element(element, parent);
+      return listeners.element && listeners.element(element, parent);
     },
     root: function(root) {
-      listeners.root && listeners.root(root);
+      return listeners.root && listeners.root(root);
     },
     error: function(message, context) {
-      listeners.error && listeners.error(message, context);
+      return listeners.error && listeners.error(message, context);
```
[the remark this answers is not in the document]
marstamm: To mimic the behavior of the Importer. We use the return values in the tree walker to establish the parent chain:
https://github.com/bpmn-io/bpmn-js/blob/master/lib/import/Importer.js#L52-L63
https://github.com/bpmn-io/bpmn-js/blob/master/lib/import/BpmnTreeWalker.js#L83

test/spec/import/BpmnTreeWalkerSpec.js
```diff
@@ -50,6 +51,35 @@ describe('import - BpmnTreeWalker', function() {
...
+    return createModdle(collaboration).then(function(result) {
+
+      var definitions = result.rootElement;
+
+      // when
+      walker.handleDefinitions(definitions);
+
+      // then
+      expect(elementSpy.neverCalledWith(sinon.match.any, sinon.match.typeOf('undefined'))).to.be.true;
+      expect(errorSpy.notCalled).to.be.true;
```
nikku, line 78:
```suggestion
      expect(errorSpy).not.to.be.called;

```

marstamm: :+1: 

Will change it with the other tests in this file as well

marstamm:
Fixes applied

nikku (approved):
:clap:

PR #1481, opened 2021-08-30
1474 support multi diagrams

marstamm:


marstamm:
(adjusted commit message to close related issue)

marstamm:
@nikku and @philippfromme ,

I added you because you already have Context on the topic. #1487 is based on this PR

lib/features/auto-resize/BpmnAutoResizeProvider.js
```diff
@@ -32,8 +32,11 @@ BpmnAutoResizeProvider.$inject = [
  * @return {boolean}
  */
 BpmnAutoResizeProvider.prototype.canResize = function(elements, target) {
-
-  if (!is(target, 'bpmn:Participant') && !is(target, 'bpmn:Lane') && !(is(target, 'bpmn:SubProcess'))) {
+  if (
```
philippfromme, line 35: How is this related to the import? Isn't the scope of this pull request the import of diagrams? I assume we're going to tackle modeling in another pull request.

marstamm: That is correct, good catch. We should separate PRs to one topic, even when the "fix" to make modeling work is simple :+1:

lib/import/BpmnImporter.js
```diff
@@ -108,7 +108,16 @@ BpmnImporter.prototype.add = function(semantic, di, parentElement) {
...
+    if (is(semantic, 'bpmn:SubProcess')) {
+      element.id = element.id + '_plane';
+    }
+
+    var plane = this._canvas.getPlane(semantic.id);
+    if (plane) {
+      this._canvas.setRootElementForPlane(element, plane);
+    } else {
+      this._canvas.createPlane(semantic.id, element);
+    }
```
marstamm, line 120: This got redundant with https://github.com/bpmn-io/diagram-js/pull/569, I will remove it

philippfromme:
I performed a test with some of our extensions. Both the minimap and the token simulation are broken by the changes.

### Minimap

The minimap adds all elements that are added to the canvas, regardless of the layer. It's not aware of layers, obviously.

chrome_CWfUNCRd8g

### Token Simulation

The token simulation accesses the DI through the business object to preserve the colors of elements.

chrome_iDkD3JnXzM

nikku:
@philippfromme Thanks for investigating the compatibility topic.

Which libraries did you try? And if this is a bigger topic (requiring work in multiple libraries), could you extract your findings into a separate issue? This way we can better track the remaining work to do.

lib/import/BpmnImporter.js
```diff
@@ -108,7 +108,11 @@ BpmnImporter.prototype.add = function(semantic, di, parentElement) {
     // add a virtual element (not being drawn)
     element = this._elementFactory.createRoot(elementData(semantic, di));
 
-    this._canvas.setRootElement(element);
+    if (is(semantic, 'bpmn:SubProcess')) {
+      element.id = element.id + '_plane';
```
philippfromme, line 112: Is the following case possible?

```xml
  <bpmndi:BPMNPlane id="BPMNPlane_1" bpmnElement="Process_1rjrv55">
    <bpmndi:BPMNShape id="Activity_1_di" bpmnElement="Subprocess" isExpanded="false">
      <dc:Bounds x="160" y="77" width="100" height="80" />
    </bpmndi:BPMNShape>
    <bpmndi:BPMNShape id="Activity_2_di" bpmnElement="Task_A">
      <dc:Bounds x="290" y="77" width="100" height="80" />
    </bpmndi:BPMNShape>
  </bpmndi:BPMNPlane>
</bpmndi:BPMNDiagram>
<bpmndi:BPMNDiagram id="BPMNDiagram_2">
  <bpmndi:BPMNPlane id="BPMNPlane_2" bpmnElement="Subprocess">
    <bpmndi:BPMNShape id="Activity_3_di" bpmnElement="Task_B">
      <dc:Bounds x="420" y="77" width="100" height="80" />
    </bpmndi:BPMNShape>
  </bpmndi:BPMNPlane>
</bpmndi:BPMNDiagram>
<bpmndi:BPMNDiagram id="BPMNDiagram_3">
  <bpmndi:BPMNPlane id="BPMNPlane_3" bpmnElement="Subprocess">
    <bpmndi:BPMNShape id="Activity_4_di" bpmnElement="Task_B">
      <dc:Bounds x="420" y="77" width="100" height="80" />
    </bpmndi:BPMNShape>
  </bpmndi:BPMNPlane>
</bpmndi:BPMNDiagram>
```

Using the `ids` utility would make sure that there are no ID clashes.

philippfromme: Generally, this is the first time we don't always have the same ID on the shape and the business object. I'm trying to think of potential issues this could cause but haven't found anything so far. 👍🏻

lib/import/BpmnImporter.js
```diff
@@ -172,7 +176,7 @@ BpmnImporter.prototype.add = function(semantic, di, parentElement) {
       // are rendered correctly across different "hacks" people
       // love to model such as cross participant / sub process
       // associations
-      parentElement = null;
+      parentElement = this._canvas.findPlane(parentElement).rootElement;
```
philippfromme, line 179: I'd love to have just a little more test coverage. This line isn't covered at all. I don't know why this change was necessary.

marstamm: I will add more test coverage for this case. This change is necessary as elements with unset `rootElement` get placed in the currently active plane. As we set the active plane after the import, the Association would be on the `base` plane instead of `rootProcessId`.

lib/import/BpmnImporter.js
marstamm: Having multiple Planes for the same Subprocess has implications beyond the ID clash, as it is no longer clear which one should be linked to the Subprocess Navigation. I suggest we don't support this configuration for now.

As for `ids` util: I tried `moddle.ids.nextPrefixed('Plane', element)`, but the `ids` utility is only present in the modeler: https://github.com/bpmn-io/bpmn-js/blob/157aec6ed6e2c5bfde10c6176473d5eda6dad1b9/lib/BaseModeler.js#L50

Is there another Utility that I could use? Creating a separate Ids instance just for import seems like bad practice.

philippfromme: Okay, got it. We don't have to tackle this right now but we might just have to introduce an IDs util for the viewer in the future. We actually use a separate IDs util in the properties panel for example: https://github.com/bpmn-io/bpmn-properties-panel/blob/main/src/utils/ElementUtil.js#L28

lib/import/BpmnImporter.js
philippfromme: Got it. 👍🏻

lib/import/BpmnImporter.js
marstamm: Should I create a follow-up issue to move the ID util into the baseViewer?

PR #1497, opened 2021-10-04
fix(path-map): disallow nested tokens

barmac:
The previous regex is risky as it can have exponential execution time if we nest tokens, e.g. `{{{{{{{`.

Still, it's not a vulnerability as we use only hard-coded values for strings passed to `format`.


philippfromme (approved):
Very nice fix! We should apply the same fix to dmn-js, too. 👏🏻

PR #1507, opened 2021-10-19
chore(bpmn-replace): reorganize toggling collapsed state

barmac:
* `toggeling` was renamed to be less cryptic
* side effect was removed
* nesting conditions were flattened


nikku:
I've also integration tested this one :muscle:.

PR #1517, opened 2021-11-03
chore: prevent double bundling

nikku:
For a long time I was wondering why bundling got so slow, especially noticeable after `npm install`. 

With the introduction proper bundle logs (https://github.com/bpmn-io/bpmn-js/commit/3dfc1dd07e91de9575bdc6339ee76dbf614338ce) the issue got apparent: We do bundle _all_ distributions twice, in `NODE_ENV=(development,production)`.

This PR ensures we build the distribution only once:

* development is built with NODE_ENV=development
* production is built with NODE_ENV=production

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -45,7 +45,7 @@ const configs = distros.reduce(function(configs, distro) {
       },
       plugins: pgl([
         banner(output)
-      ])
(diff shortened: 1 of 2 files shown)


barmac (approved):
giphy

PR #1519, opened 2021-11-04
style: update used colors

pinussilvestrus:
Related to https://github.com/bpmn-io/diagram-js/issues/581
Requires https://github.com/bpmn-io/diagram-js/pull/583 to be merged and integrated.

This only effects the subprocess navigation features, in a minimal fashion:

_Before_
image

_After_
image

### DIFF
--- a/assets/bpmn-js.css
+++ b/assets/bpmn-js.css
@@ -2,6 +2,13 @@
   --bjs-font-family: Arial, sans-serif;
 }
 
+.djs-container {
+  --breadcrumbs-item-color: var(--color-blue-205-100-50);
+  --breadcrumbs-arrow-color: var(--color-black);
+  --drilldown-fill-color: var(--color-white);
+  --drilldown-background-color: var(--color-blue-205-100-50);
+}
+
 .bjs-breadcrumbs {
   position: absolute;
   display: flex;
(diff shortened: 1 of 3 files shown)


assets/bpmn-js.css
```diff
@@ -2,6 +2,13 @@
   --bjs-font-family: Arial, sans-serif;
 }
 
+.djs-container {
+  --breadcrumbs-item-color: var(--color-blue-205-100-50);
+  --breadcrumbs-arrow-color: var(--color-black);
+  --drilldown-fill-color: var(--color-white);
+  --drilldown-background-color: var(--color-blue-205-100-50);
```
marstamm, line 9: as these are only bpmn-js releated, can we move them to `.bjs-container`?

pinussilvestrus: Not really, because the `.bjs-container` is a parent of the `.djs-container`

image

The full color scheme is defined in the `.djs-container`, therefore they wouldn't be available in the `.bjs-container`.

pinussilvestrus: An alternative would be to define the colors again in the `.bjs-container`, but I'm not sure whether this makes sense.

marstamm: Ah, makes sense :+1:

marstamm (approved):

PR #1521, opened 2021-11-04
Drilldown documentation and fixes

marstamm:
This PR backports mutliple smaller fixes from https://github.com/bpmn-io/bpmn-js/pull/1506:

- Fix the moddle of legacy processes before import
- rename `subprocess-navigation` to `drilldown` and add documentation
- center breadcrumb arrow


nikku (commented):
@marstamm Thanks for collecting all our downstream fixes into this PR.

What is best regarding conventional commits is that we use the modules we got for scoping (`modeling`, `keyboard`, `core`). Some examples: 

* Rather than referring to `chore(styling)` refer to the module that is being worked on, if non exists (general refresh), simply use `chore: cleanup this and that`.
* Similar to `fix(import):`. This clearly belongs to a feature called `subprocess-compatibility`, right`? I'd propose to move it into that self-contained feature and use `subprocess-compatibility` or however you call it as a scoping label.
* `fix(css)`: which component is broken here and thus fixed? It is the subprocess-navigation component, right (or `drilldown` as we renamed it to)? The conventional commit scope shall reflect that.
Could

marstamm:
Thank you for the hints, that makes sense. Let's scope it as `drilldown`, as we also renamed the module to make that clearer

nikku (approved):

PR #1531, opened 2021-11-15
chore(drilldown): make CSS more robust

### LINKED ISSUE #1529: Breadcrumb CSS easily breaks in integrating applications
__Describe the Bug__

The Breadcrumbs don't look like they are supposed to when integrated in Cockpit. We should consider making the CSS more robust against external factors, even when they are not needed all the time.

image

Note: 
- Arrows and text are not aligned
- Last item has link color

__Steps to Reproduce__

1. have global `line-height` set
2. have global color of links set
3. it affects the breadcrumbs

https://codesandbox.io/s/simple-bpmn-js-example-forked-4bqge

__Expected Behavior__

- line height does not mess with arrow positioning
- Last crumb always uses default text font

__Environment__

 - Browser: Chrome 90
 - OS: Linux
 - Library version: bpmn-js@9.0.0-subprocesses.2

marstamm:
closes #1529

**How to test**
In `test/TestHelper.js`, you can add css for the test cases:

```javascript
insertCSS('foobar.css', `
body {
  line-height: 3px;
}

a {
  color: green;
}
`);
```

### DIFF
--- a/assets/bpmn-js.css
+++ b/assets/bpmn-js.css
@@ -13,21 +13,25 @@
   position: absolute;
   display: flex;
   flex-wrap: wrap;
+  align-items: center;
   top: 10px;
   left: 100px;
   font-family: var(--bjs-font-family);
   font-size: 16px;
-  align-items: center;
+  line-height: normal;
 }
 
 .bjs-breadcrumbs li {
   display: inline-flex;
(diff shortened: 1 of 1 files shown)


pinussilvestrus (approved):
LGTM

PR #1535, opened 2021-11-23
chore(css): expose color variables in bpmn-js.css

marstamm:
Allows the use of planes in the viewer without including `diagram-js.css`

### DIFF
--- a/assets/bpmn-js.css
+++ b/assets/bpmn-js.css
@@ -1,8 +1,34 @@
 .bjs-container {
   --bjs-font-family: Arial, sans-serif;
-}
 
-.djs-container {
+  --color-grey-225-10-15: hsl(225, 10%, 15%);
+  --color-grey-225-10-35: hsl(225, 10%, 35%);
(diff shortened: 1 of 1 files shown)


assets/bpmn-js.css
```diff
@@ -1,8 +1,34 @@
 .bjs-container {
   --bjs-font-family: Arial, sans-serif;
-}
 
-.djs-container {
+  --color-grey-225-10-15: hsl(225, 10%, 15%);
```
nikku, line 4: We do only need one color, correct? Why would we include all of themn?

marstamm: Correct, we only need certain colors (blue, black, white). I wanted to keep it consistent with diagram-js, where we also define colors we don't use (e.g. [`--color-grey-225-10-90`](https://github.com/bpmn-io/diagram-js/blob/develop/assets/diagram-js.css#L11)).

nikku: In this case, O.K. I guess. At least we're consistent.

PR #1549, opened 2021-12-07
feat(modeling): don't select start-event inside newly created sub-process

nikku:
The canonical modeling operation shall not be drill down, but continue to model.

capture 4a9kuq_optimized


andreasgeier (approved):
Makes sense. Approved.

MaxTru (approved):

barmac:
Makes sense ✅

Issue #1546, opened 2021-12-06
Ensure compatibility with single-plane imports


marstamm:
- Single imports were a compromise for multi-diagram files when we did not support multi-plane imports
- We can now import multiple diDiagrams at a time
- Instead of "single diagram", `viewer.open()` should just switch to the corresponding plane

- When a Diagram has multiple DI representations, ensure that only one is loaded

PR #1551, opened 2021-12-09
Use multi-root aware diagram-js

nikku:
Related to https://github.com/bpmn-io/diagram-js/pull/601.

### DIFF
--- a/lib/features/drilldown/DrilldownOverlays.js
+++ b/lib/features/drilldown/DrilldownOverlays.js
@@ -7,6 +7,7 @@ var ARROW_DOWN_SVG = '<svg xmlns="http://www.w3.org/2000/svg" width="20" height=
 
 var OPEN_CLASS = 'bjs-breadcrumbs-shown';
 
+
 /**
  * Adds Overlays that allow switching planes on collapsed subprocesses.
  *
--- a/lib/features/drilldown/DrilldownOverlays.js
+++ b/lib/features/drilldown/DrilldownOverlays.js
@@ -21,21 +21,15 @@ export default function DrilldownOverlays(eventBus, elementRegistry, overlays, c
   var containerClasses = classes(container);
   container.appendChild(breadcrumbs);
 
-  function updateBreadcrumbs(plane) {
-    var subProcess = elementRegistry.get(plane.name);
-    var parents = getParentChain(subProcess);
+  function updateBreadcrumbs(element) {
+    var parents = getParentChain(element);
 
     var path = parents.map(function(el) {
       var title = escapeHTML(el.name || el.id);
       var link = domify('<li><span class="bjs-crumb"><a title="' + title + '">' + title + '</a></span></li>');
 
       link.addEventListener('click', function() {
-        if (canvas.getPlane(el.id)) {
-          canvas.setActivePlane(el.id);
-        } else {
-          var plane = canvas.findPlane(el.id);
-          canvas.setActivePlane(plane);
-        }
+        canvas.setRootElement(canvas.findRoot(planeId(el)) || canvas.findRoot(el.id));
       });
 
       return link;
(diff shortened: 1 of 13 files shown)


nikku:
@marstamm To be examined.

lib/import/Importer.js
```diff
@@ -76,7 +76,12 @@ export function importBpmnDiagram(diagram, definitions, bpmnDiagram) {
     }
 
     var mainDiagram = bpmnDiagram || definitions.diagrams[0];
-    canvas.setActivePlane(mainDiagram.plane.bpmnElement.id);
+
+    canvas.setRootElement(
```
nikku, line 80: Question to you @marstamm: How do we currently import a process + drill down right away?

I guess we don't support case, but rather import a single diagram only when it is being passed to the bpmn-js instance via `#open`:

```
await bpmnJs.open(diagramXML, 'ID_OF_DIAGRAM');
```

nikku: If we did support importing all + opening a single diagram, we'd need to change this line below :arrow_down:.

marstamm: No, we do not support it at this point. However, it is on our roadmap: https://github.com/bpmn-io/bpmn-js/issues/1546

nikku: In this case my proposal is to make importing future proof already: https://github.com/bpmn-io/bpmn-js/pull/1551/commits/67f42f53492c416ebf282cf80e95bce5e3002d67

nikku:
Pending https://github.com/bpmn-io/diagram-js/pull/601 to be asserted and merged.

lib/features/drilldown/DrilldownOverlays.js
```diff
@@ -107,4 +101,13 @@ var getParentChain = function(child) {
...
+}
+
+
+function planeId(element) {
+  if (is(element, 'bpmn:SubProcess')) {
+    return element.id + '_plane';
+  }
+
+  return element.id;
+}
```
marstamm, line 113: Would it make sense to expose this function in a utility Function? We will probably need it in other places as well.

lib/import/Importer.js
```diff
@@ -76,7 +76,15 @@ export function importBpmnDiagram(diagram, definitions, bpmnDiagram) {
...
     var mainDiagram = bpmnDiagram || definitions.diagrams[0];
-    canvas.setActivePlane(mainDiagram.plane.bpmnElement.id);
+
+    var rootId = mainDiagram.plane.bpmnElement.id;
+
+    // we do need to account for different ways we create root elements
+    // each nested imported <root> do have the `_plane` suffix, while
+    // the root <root> is found under the business object ID
+    canvas.setRootElement(
+      canvas.findRoot(rootId + '_plane') || canvas.findRoot(rootId)
```
marstamm, line 86: Here we could use the `planeId` utility

lib/features/drilldown/DrilldownOverlays.js
nikku: We _could_ expose it. However, this is not a safe operation _per se_. If we render the sub-process as a root (only element imported) this naming assumption does not longer hold. 

TLDR: I'd love to continue to improve this in the future, if we have the full picture, not now.

lib/import/Importer.js
nikku: Unfortunately we cannot use the utility here. How to find the root depends on whether we import the sub-process as the root of the diagram (it is called `subprocess(nosuffix)` then or as a nested sub-process plane, called `subprocess_plane` then).

nikku: => It is safer to look it up searching for plane first, and actual element next.

lib/features/drilldown/DrilldownOverlays.js
marstamm: Makes sense. Lets not do it now and expose it when we need it :+1:

lib/import/Importer.js
marstamm: Makes sense with the current semantic of `open`. 
We should revisit this when we tackle https://github.com/bpmn-io/bpmn-js/issues/1546, as "subProcess is root only" should not exist after that

nikku:
Please give your approval. We can iterate on the details in the course of time :ship:

marstamm (approved):

PR #1552, opened 2021-12-11
feat(util/ModelUtil): expose `isAny`

nikku:
It is a core utility that will be tree-shaken if not used.

We'll continue to expose it via `ModelingUtil` for backwards
compatibility.


barmac (approved):
Cool improvement :) It was always confusing to import one thing from the ModelUtil and another from the ModelingUtil.

nikku:
Changed the base to `multi-root` as our test suite will otherwise explode.

PR #1554, opened 2021-12-15
fix(bpmnReplace): ensure `collapsed` is set before replacing

marstamm:
With the refactoring in e1efb556f2cb7cfb92a2361b9dfa0ea96dbadebd, we broke the following scenario:

- Create a collapsed subprocess
- Try to expand it

We need to ensure the `collapsed` property is set before expanding/collapsing the shape.

### DIFF
--- a/lib/features/modeling/ElementFactory.js
+++ b/lib/features/modeling/ElementFactory.js
@@ -117,6 +117,10 @@ ElementFactory.prototype.createBpmnElement = function(elementType, attrs) {
     applyAttribute(di, attrs, 'isExpanded');
   }
 
+  if (is(businessObject, 'bpmn:SubProcess')) {
+    attrs.collapsed = !isExpanded(businessObject, di);
+  }
+
   if (is(businessObject, 'bpmn:ExclusiveGateway')) {
     di.isMarkerVisible = true;
   }
--- a/test/spec/features/modeling/ElementFactorySpec.js
+++ b/test/spec/features/modeling/ElementFactorySpec.js
@@ -116,6 +116,19 @@ describe('features - element factory', function() {
     }));
 
 
+    it('should add collapsed attribute to subprocess', inject(function(elementFactory) {
+
+      // when
+      var subprocess = elementFactory.createShape({
+        type: 'bpmn:SubProcess',
+        isExpanded: false
+      });
+
+      // then
+      expect(subprocess.collapsed).to.be.true;
+    }));
+
+
     describe('integration', function() {
 
       it('should create event definition with ID', inject(function(elementFactory) {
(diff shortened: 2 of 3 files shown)


lib/features/replace/BpmnReplace.js
```diff
@@ -108,6 +121,7 @@ export default function BpmnReplace(
 
     if (isSubProcess(oldBusinessObject) && type === 'bpmn:SubProcess') {
       if (shouldToggleCollapsed(element, target)) {
+        ensureCollapsedProperty(element);
```
nikku, line 124: Why could it not be set in the first place?

nikku: I'd rather fix it over there (i.e. assume it is always appropriately set, or falsy).

marstamm: You mean in the `elementFactory`? This could work.

Do we always use `elementFactory` when modifying shapes? I'm asking because I want want to make sure we don't have a scenario where we can have a shape that was morphed into a subProcess and therefore does not have a valid collapsed attribute

nikku: I don't mean `ElementFactory`. I literally mean the one place where the shape is being created. Likely it is the element factory though :slightly_smiling_face:.

From the architecture perspective (bpmn-js wise) we can assume that shapes are _always_ created via the `ElementFactory` before being added to the diagram. If it is not the case the user is miss-using our system.

nikku: From the approach point of view, let us try to favor close-to-root-cause fixes over than monkey-patched-somewhere.

marstamm: Can we safely assume that the type of the shape does not change? E.g., a Task would have `collapsed = undefined`, which would mean an expanded shape if it is somehow morphed into a subprocess

nikku: Thought: `collapsed` is legacy. Maybe `expanded` is a better attribute to describe what we want to do here? task natually would be `expanded = undefined == false`.

nikku: Morphing to a shape of a different type is _always_ a _replace_ operation.

Replacing implies adding an element to the editor via `Modeling#...`. This implies [creating a new element via the elementFactory](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/modeling/Modeling.js#L506).

Does this answer your question?

barmac: Thought: Never use "collapsed" property and always resolve to the `BPMNShape#isExpanded`.

marstamm: As discussed with Nico:

- We use `collapsed` also in `diagram-js`, so we can't easily get rid of it
- We always create our Shapes in the `ElementFactory`, it is safe to set it there

--> Let's ensure it is present in the ElementFactory

nikku (approved):
Very nice!

PR #1559, opened 2022-01-06
fix(tasks): skip wiredeps per default

### LINKED ISSUE #1558: Builds based on `master` are broken due to incompatible diagram-js version
__Describe the Bug__

Our CI on branches targeting `master` is currently broken, cf. [this build](https://github.com/bpmn-io/bpmn-js/runs/4725608553?check_suite_focus=true#step:6:18).

It's working fine locally though.

__Steps to Reproduce__

1. Checkout [this build](https://github.com/bpmn-io/bpmn-js/runs/4725608553?check_suite_focus=true#step:6:18)
2. See the build is failing on 25 broken test cases

__Expected Behavior__

Make CI work again.

__Environment__

 - Browser: not specified
 - OS: discovered on Linux & Mac, Windows seem to work fine.
 - Library version: `master`

pinussilvestrus:
Closes #1558 

Proof it works: https://github.com/bpmn-io/bpmn-js/runs/4726762563

### DIFF
--- a/tasks/wiredeps
+++ b/tasks/wiredeps
@@ -17,7 +17,12 @@ echo "Attempting to install diagram-js@$FEATURE_BRANCH";
 npm install "diagram-js@bpmn-io/diagram-js#$FEATURE_BRANCH";
 
 if [ $? -ne 0 ]; then
-  echo "Falling back to diagram-js@develop";
+   
+  # Skipping the fallback due to incompatible master branch
+  # https://github.com/bpmn-io/bpmn-js/issues/1558
+  exit 0;
 
-  npm install "diagram-js@bpmn-io/diagram-js#develop";
+  #echo "Falling back to diagram-js@develop";
+
+  #npm install "diagram-js@bpmn-io/diagram-js#develop";
 fi


nikku:
Alternative to consider: Change the default branch to build against to master; we'll need to undo this change once bpmn-js@next-major-alpha is released; which is some day, soon.

pinussilvestrus:
Thanks for your thoughts! I agree that this flag might also be just an interim solution and may produce more overhead than needed.

I'm not sure whether defaulting to `diagram-js#master` would solve the issue, since the API changes are already released there as well, which would still cause the failing tests for branches targeting `bpmn-js#master`. 

What about to simply (temporarily) skipping `wiredeps` if installing the upstream branch fails and simply using the diagram-js version which is defined in the package.json? We could revert this change afterward.

pinussilvestrus:
I sketched this here: https://github.com/bpmn-io/bpmn-js/pull/1559/commits/f35ccdd3f8bfc19d22334d4016d6ae5cd369cbb4

nikku:
> I'm not sure whether defaulting to diagram-js#master would solve the issue, since the API changes are already released there as well, which would still cause the failing tests for branches targeting bpmn-js#master.

The point would be to make bpmn-js master not failing anymore, i.e. release a breaking alpha release on it. In cases of urgent backports we usually create prev-major branches to do that there.

pinussilvestrus:
We decided to follow up with an alpha release to get things in sync again. @marstamm will take it over 👍 

I'll close this one here, once we see this situation coming more often, we can decide to change things in our branching strategy.

PR #1563, opened 2022-01-10
feat(copy-paste): allow copy of subprocess planes

### LINKED ISSUE #1556: Copy-Paste multiple layers

marstamm:
closes #1556

During Copy/Paste, we handle the subprocess as if it is on a single plane. When Pasting, the children are then moved to the correct new planes.


philippfromme (approved):
🐎

PR #1573, opened 2022-01-13
Backport fixes to `8.x`

nikku:
This backports a couple of fixes to the `8.x` series of our library and prepares a 8.9.0 release.

PR #1574, opened 2022-01-13
chore: skip wiredeps on backport branches

nikku:
This extends our wiredeps skip logic to not only match `master` branches and tags but also any backport branches `\d.x` as well as PRs that target master and backport branches.

Why: Because we want to safely backport things (and fix things) without wiredeps interferring.

---

image

---

Proof it works: https://github.com/bpmn-io/bpmn-js/pull/1573.


pinussilvestrus (approved):
Cool stuff!

PR #1579, opened 2022-01-18
fix(modeling): ensure di for embedded labels

### LINKED ISSUE #1540: `bpmndi:BPMNLabel` is not set on `bpmndi:BPMNShape`s for internal labels
__Describe the Bug__

We do not set `bpmndi:BPMNLabel` on a `bpmndi:BPMNShape` when the label is not external (movable), although it's required by the spec.

[On page 411 of the BPMN 2.0 spec](https://www.omg.org/spec/BPMN/2.0/PDF#page=411), it's specified that:

> Simply put, label visibility is defined by the presence of the BPMNLabel element.

Which means that we should set the label if we want to perform diagram interchange with other vendors correctly. The bounds are not required.

It's currently causing a problem with Software AG tool only AFAIK.

__Steps to Reproduce__

1. Create a task and give it a name.
2. Check the XML part

```xml
<bpmndi:BPMNShape id="Activity_06dr0no_di" bpmnElement="Activity_06dr0no">
  <dc:Bounds x="300" y="340" width="100" height="80" />
</bpmndi:BPMNShape>
```

__Expected Behavior__

It should serialize to:

```diff
<bpmndi:BPMNShape id="Activity_06dr0no_di" bpmnElement="Activity_06dr0no">
  <dc:Bounds x="300" y="340" width="100" height="80" />
+ <bpmndi:BPMNLabel />
</bpmndi:BPMNShape>
```

__Environment__

Any version up to `v8.8.2`.

---

Related to https://github.com/bpmn-io/bpmn-js/issues/1537

Discovered during a BPMN MIWG session.

marstamm:
closes #1540

### DIFF
--- a/lib/features/label-editing/cmd/UpdateLabelHandler.js
+++ b/lib/features/label-editing/cmd/UpdateLabelHandler.js
@@ -11,6 +11,7 @@ import {
 } from '../../../util/LabelUtil';
 
 import {
+  getDi,
   is
 } from '../../../util/ModelUtil';
 
@@ -23,7 +24,30 @@ var NULL_DIMENSIONS = {
 /**
  * A handler that updates the text of a BPMN element.
  */
-export default function UpdateLabelHandler(modeling, textRenderer) {
+export default function UpdateLabelHandler(modeling, textRenderer, bpmnFactory) {
+
+  /**
+   * Creates an empty `diLabel` attribute for embedded labels.
+   *
+   * @param {djs.model.Base} element
+   * @param {string} text
+   */
+  function ensureInternalLabelDi(element, text) {
+    if (isLabelExternal(element)) {
+      return;
+    }
+
+    var di = getDi(element);
+
+    if (text && !di.label) {
+      di.label = bpmnFactory.create('bpmndi:BPMNLabel');
+    }
+
+    if (!text && di.label) {
+      di.label = null;
+    }
+  }
+
 
   /**
    * Set the label and return the changed elements.
@@ -41,6 +65,8 @@ export default function UpdateLabelHandler(modeling, textRenderer) {
     var labelTarget = element.labelTarget || element;
 
     setLabel(label, text, labelTarget !== label);
+
+    ensureInternalLabelDi(element, text);
 
     return [ label, labelTarget ];
   }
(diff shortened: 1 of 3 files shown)


pinussilvestrus (approved):
👏 One step further!

barmac:
Well done! 👏

lib/features/label-editing/cmd/UpdateLabelHandler.js
```diff
@@ -23,7 +24,30 @@ var NULL_DIMENSIONS = {
...
+   */
+  function ensureInternalLabelDi(element, text) {
+    if (isLabelExternal(element)) {
+      return;
+    }
+
+    var di = getDi(element);
+
+    if (text && !di.label) {
+      di.label = bpmnFactory.create('bpmndi:BPMNLabel');
```
nikku, line 43: What I'm wondering here: How much more valid is a label without bounds? Did we really advance our BPMN support here?

Proposal: Set the label width/height/x/y to the element bounds (or position for participant / sub-process).

nikku: My proposal obviously depends on what the BPMN spec says. How do other vendors handle this, what do they expect in terms of embedded DI labels?

marstamm: This is actually defined in the spec:

> The depiction resolution tables provided below exemplify default label positions for
BPMNShape kinds (Section 12.3.2) and BPMNEdge kinds (Section 12.3.3) if no BPMNLabel bounds are provided.

The default position for Labels without bounds is the middle of the shape

marstamm: cf. https://www.omg.org/spec/BPMN/2.0/PDF#page=411

barmac: @nikku Check out the issue: https://github.com/bpmn-io/bpmn-js/issues/1540

nikku: Thanks for clarifying. You're awesome folks :tada:!

PR #1585, opened 2022-01-25
chore(draw): expose extension points

nikku:
+ clearly documented as *UNSAFE*
+ add test cases
+ tested existing *UNSAFE* handler endpoint

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -382,14 +382,14 @@ export default function BpmnRenderer(
     return drawPath(parentGfx, path, assign({ 'data-marker': type }, attrs));
   }
 
-  function as(type) {
-    return function(parentGfx, element) {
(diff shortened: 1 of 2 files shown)


marstamm:
Could you give a little context on the use-case? I guess it is to get access to the render functions outside of the `BpmnRender`, correct?

nikku:
Exactly. The use-case is that I am plugging in my own renderer to do some work, but delegate to the `bpmnRenderer` for the heavy lifting.

I.e. I want to render a task but add my special icon to it.

marstamm (approved):
:rocket:

nikku:
The way to do this is to check internal state (cf. https://github.com/bpmn-io/connectors-modeling-demo/blob/master/connectors-extension/ConnectorTaskRenderer.js#L56).

By having an indirection this becomes easier to handle.

PR #1590, opened 2022-02-01
test(drilldown): add integration test for multi-di planes

marstamm:
related to https://github.com/bpmn-io/bpmn-js/pull/1576#discussion_r795532712


nikku:
Very nice!

PR #1612, opened 2022-03-01
fix(devSetup): build bpmn-moddle in context of devSetup

MaxTru:
* This will ensure that `npm run start` actually works after executing the setup script

### DIFF
--- a/docs/project/setup.sh
+++ b/docs/project/setup.sh
@@ -26,6 +26,8 @@ echo setup bpmn-moddle
 cd $base/bpmn-moddle
 npm install
 
+npm run build
+
 
 echo setup bpmn-js
 


docs/project/setup.sh
```diff
@@ -26,6 +26,8 @@ echo setup bpmn-moddle
 cd $base/bpmn-moddle
 npm install
 
+npm run build
```
nikku, line 29: Could also be fixed by changing from `prepublishOnly` to `prepare` [here](https://github.com/bpmn-io/bpmn-moddle/blob/master/package.json#L13).

nikku: (That would make `bpmn-moddle` `npm link` friendly).

MaxTru: true, we also discused this in a weekly. Will follow-up

Issue #1619, opened 2022-03-16
Sub process doesn't render dotted lines despite that they exist in the XML file.


3ronco:
Although rather uncommon we use a subprocess node with the ability to give 'documents' or 'parameters' to them via a _DataObjectReference_ node. Earlier versions showed these flows as a dotted line among a document and a subprocess.
The recent version doesn't render these arrows anymore although they can be found as shape nodes in the XML file. It's not happening with earlier versions eg. _V8.7.1_

__Reproduce__
1. Goto https://demo.bpmn.io/new
2. **Append task** to start event (optionally give it some name)
3. Switch task to type **Sub Process (collapsed)**.
4. Drag a new **DataObjectReference** above the _sub process_.
5. Wire **DataObjectReference** with (dotted arrow) flow to _sub process_.
6. Drag a new **DataObjectReference** below the _sub process_.
7.  Wire the _sub process_ with a outgoing flow to the **DataObjectReference** below.
8. Click _"Download as BPMN 2.0 file"_ ...
9. Create a new tab ... https://demo.bpmn.io/new
10. Click _"Open BPMN diagram file from local file system"_ and select the previously downloaded file.
11. The file loads but without dotted arrows among **DataObjectReference** nodes and _sub processes_ (and only among these types!)

__Expected Behavior__
The recently opened diagram file renders exactly the same way as it was when modeled.

__Environment__
 - Browser: _Firefox 97.0.2_
 - OS: _Windows 10_
 - Library version: _most recent_

marstamm:
Hi @3ronco ,

thank you for reporting this bug. I was able to reproduce and verify the bug.

## Root Cause

The dataInput and dataOutput associations are children of the Activity, where sequence flows are part of the Process.

related BPMN Spec:
![image](https://user-images.githubusercontent.com/21984219/158606674-b4a87dff-62dc-4271-893a-890171a30129.png)

With bpmn-js@9, we move all children of subprocesses into the subprocess plane.
https://github.com/bpmn-io/bpmn-js/blob/bdf9cf3e752254a4c8172031d8a493955a9fca9c/lib/features/drilldown/SubprocessCompatibility.js#L82-L86

We did not account for this edge-case, where the element is a child of the subprocess but displayed on root.

## Fix
Add a check for edge-cases to the code above to ensure we don't move the data associations into the wrong plane

PR #1620, opened 2022-03-16
render dataAssociations on correct parent

marstamm:
From what I got from the bpmn documentation, the Input/Output Associations are the only edge cases where the element is a child of the subprocess but rendered outside of it. Please correct me if I am missing something!

image

closes https://github.com/bpmn-io/bpmn-js/issues/1619


nikku (approved):
Looks good to me.

PR #1626, opened 2022-03-29
Add DOM style attributes in CSP-compliant way

### LINKED ISSUE #1625: BPMN viewer conflicts with strict Content-Security-Policy (CSP) style-src directive
__Describe the Bug__

  

__Steps to Reproduce__

1. Run the BPMN viewer within a web application which has strict `style-src` directives in its Content Security Policy (CSP) header

__Expected Behavior__

Style manipulations should be performed in a way which does not break CSP rules.

__More Information__

See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src

> the following stylesheets are blocked and won't load [...] Inline style attributes are also blocked [...] As well as styles that are applied in JavaScript by setting the style attribute directly, or by setting `cssText`

The problematic parts of the BPMN code relate to uses of the `domify` helper. When used with inline styles, a strict `style-src` directive will block the styles from taking effect. The styles should be `set directly on the element's style property`, as shown in the above MDN link.

__Environment__

 - Browser: Any of these: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src#browser_compatibility
 - OS: Any
 - Library version: 6.0.0 and up

----

Also reported via [internal conversation](https://camunda.slack.com/archives/CKGH9LR40/p1651476039251279).

WSydnA:
Closes #1625

### DIFF
--- a/lib/BaseViewer.js
+++ b/lib/BaseViewer.js
@@ -34,6 +34,10 @@ import {
   wrapForCompatibility
 } from './util/CompatibilityUtil';
 
+import {
+  applyStyles
+} from './import/Util';
+
 /**
  * A base viewer for BPMN 2.0 diagrams.
  *
@@ -652,14 +656,14 @@ BaseViewer.prototype._emit = function(type, event) {
 };
 
 BaseViewer.prototype._createContainer = function(options) {
+  var containerStyles = {
+    'width': ensureUnit(options.width),
+    'height': ensureUnit(options.height),
+    'position': options.position
+  };
 
   var container = domify('<div class="bjs-container"></div>');
-
-  assign(container.style, {
-    width: ensureUnit(options.width),
-    height: ensureUnit(options.height),
-    position: options.position
-  });
+  applyStyles(container, containerStyles);
 
   return container;
 };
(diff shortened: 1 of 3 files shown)


lib/BaseViewer.js
```diff
@@ -655,11 +655,9 @@ BaseViewer.prototype._createContainer = function(options) {
 
   var container = domify('<div class="bjs-container"></div>');
 
-  assign(container.style, {
-    width: ensureUnit(options.width),
-    height: ensureUnit(options.height),
-    position: options.position
-  });
+  container.style.width= ensureUnit(options.width);
```
nikku, line 658: Use `applyStyles` here.

WSydnA: I've made this change. In doing so I moved the `applyStyles` util into `lib/import/Util.js` to make it more logically accessible to other parts of the library.

nikku:
It was great if we were able to test this in our test suite, but [apparently we cannot](https://github.com/karma-runner/karma/issues/3260).

marstamm:
Hi @WSydnA ,

thank you for your work and bringing it to our attention

We decided to move the `applyStyle` utility to our [`min-dom`](https://github.com/bpmn-io/min-dom) library, so it is reusable through all components. We integrated and merged it in #1645 and I will be closing this PR.

Thanks again for your contribution.

PR #1633, opened 2022-04-11
chore(tasks): update website with published time

nikku:
Updates website with published time on release:

Cf. https://github.com/bpmn-io/bpmn.io/blob/master/src/data/site.yml#L2.


barmac:
See https://github.com/bpmn-io/dmn-js/pull/696/files#diff-7b7744b6dfce578724c9f5e005a7becb4d63933a33da463b174487a6180190c2R19-R26

barmac (approved):

PR #1645, opened 2022-05-16
fix: comply with style-src CSP

marstamm:
This PR removes all inline styles and replaces them with direct assignment to comply with CSP style-src.

closes https://github.com/bpmn-io/bpmn-js/issues/1625

### DIFF
--- a/lib/BaseViewer.js
+++ b/lib/BaseViewer.js
@@ -13,6 +13,7 @@ import {
 
 import {
   domify,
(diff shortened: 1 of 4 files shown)


nikku:
@marstamm are you aware of https://github.com/bpmn-io/bpmn-js/pull/1626? Seems to be fixing the same things, albeit without our utility. Consider merging both PRs or closing the other one with a comment.

PR #1651, opened 2022-05-18
deps: replace inherits with inherits-browser

nikku:
This increase the safety of our build; external consumers
do no longer need to account for the `browser` field to
bundle bpmn-js (or otherwise bundle a Node shim, unintentionally).

----

Related to https://github.com/bpmn-io/diagram-js/pull/639

### DIFF
--- a/lib/BaseModeler.js
+++ b/lib/BaseModeler.js
@@ -1,4 +1,4 @@
-import inherits from 'inherits';
+import inherits from 'inherits-browser';
 
 import Ids from 'ids';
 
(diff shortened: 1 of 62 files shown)


barmac:
Rebased on top of develop to fix the CI errors.

nikku:
I guess this is due to diagram-js still using the old `inherits`? Should be fixed once we incorporate a fixed diagram-js.

barmac:
Yeah, I guess so.

barmac (approved):

PR #1652, opened 2022-05-18
feat: allow TextAnnotation for SequenceFlows

marstamm:
Recording 2022-05-18 at 16 38 11

related to https://github.com/camunda/camunda-modeler/issues/402

### DIFF
--- a/lib/features/modeling/behavior/LayoutConnectionBehavior.js
+++ b/lib/features/modeling/behavior/LayoutConnectionBehavior.js
@@ -1,0 +1,117 @@
+import {
+  assign
+} from 'min-dash';
+
+import inherits from 'inherits';
+
+import CommandInterceptor from 'diagram-js/lib/command/CommandInterceptor';
+
+import { getConnectionAdjustment as getConnectionAnchorPoint } from './util/ConnectionLayoutUtil';
+
+/**
+ * A component that makes sure that Associations connected to Connections
+ * are updated together with the Connection.
+ *
+ * @param {EventBus} eventBus
+ * @param {Modeling} modeling
+ */
+export default function LayoutConnectionBehavior(
+    eventBus, modeling) {
+
+  CommandInterceptor.call(this, eventBus);
+
+  function getnewAnchorPoint(event, point) {
+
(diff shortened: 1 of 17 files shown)


lib/features/modeling/behavior/LayoutConnectionBehavior.js
```diff
@@ -0,0 +1,117 @@
...
+import { getConnectionAdjustment as getConnectionAnchorPoint } from './util/ConnectionLayoutUtil';
+
+/**
+ * A component that makes sure that Associations connected to Connections
+ * are updated together with the Connection.
+ *
+ * @param {EventBus} eventBus
+ * @param {Modeling} modeling
+ */
+export default function LayoutConnectionBehavior(
```
pinussilvestrus, line 18: Codecov reports some missing test coverage for this one. Is there anything we can do against it?

pinussilvestrus (changes requested):
I tried a bit around and discovered the positioning is not working properly in some cases

Kapture 2022-05-20 at 11 08 06

marstamm:
That's weird, It does not work for me now either. I'm 100% certain I tested this before :sweat_smile: 

I'll check if something broke during rebasing or something

lib/features/modeling/behavior/LayoutConnectionBehavior.js
marstamm: The Missing tests are for outgoing connections. Currently, text annotations always use incoming connections. I'm not sure if there is a use-case for outgoing connections, but it feels weird to only have logic for one and not the other

marstamm: I guess the bpmn spec allows associations between 2 sequence flows, but we don't support those yet

marstamm: > Currently, text annotations always use incoming connections

Disregard that, it works, we just can't model them. I'll add tests

marstamm:
Should be fixed. I messed up going from spread operators to ES5 Code :see_no_evil:

PR #1668, opened 2022-05-25
fix: use inherits-browser

nikku:
Plus add lint rule to verify it.

----

Related to https://github.com/bpmn-io/diagram-js/pull/649


barmac (approved):
Great idea with the eslint rule.

Issue #1681, opened 2022-06-09
Alignment is broken when subprocess and its child are selected


barmac:
__Describe the Bug__

https://user-images.githubusercontent.com/28307541/172854624-185a812c-1fb2-49dc-b242-f5a114ffa36f.mov

__Steps to Reproduce__

1. Select a subprocess and its child
2. Use alignment tool (via API or buttons added in your application)

__Expected Behavior__

Subprocess should not move or be resized.

__Environment__

 - Browser: [e.g. IE 11, Chrome 69]
 - OS: [e.g. Windows 7]
 - Library version: [e.g. 2.0.0]

PR #1682, opened 2022-06-10
Add alignment and distribution buttons to the context pad

### LINKED ISSUE #1680: Alignment and distribution

barmac:
Closes #1680


nikku:
What UI do we aim for? Most tooks do not show all helpers right away, but instead open an "align/distribution popup", cf.: 

capture 4rXti5_optimized

I'd suggest us to do the same.

barmac:
So this is how it would look like directly in the context pad:

image

I think it makes sense to use a popup menu instead, as it's too cluttered right now.

barmac:
I switched to popup menu-based implementation:
image

I am not satisfied with how it looks like. I'd rather aim for something closer to Miro's solution:

image

Note this does not include distribution buttons.

barmac:
With distribution buttons:
image

This is a sketch and not the current implementation state.

smbea:
I like the placement. But I think the icons feel a bit crowded, partially because being all outlines makes it harder to instantly figure out what the icon is.

In the example Nico gave, the icons appear cleaner. Also another example:

Screenshot 2022-06-24 at 11 56 47

This is just my perception. I know we are not designers and have little design resources at the moment 😅 I can try and help sketching the icons if needed.

barmac:
With more space (6px 8px):

image

barmac:
At the moment, I am mostly re-using the icons from Camunda Modeler (the one in the context pad is mine). I am open to re-designing the icons as a next step after we get a right layout.

barmac:
Cross-posting how it's done in Miro:

image

smbea:
@barmac makes sense. Looks nice for the first iteration, specially with more space. Nice work

barmac:
Just added a second sketch with even more space to https://github.com/bpmn-io/bpmn-js/pull/1682#issuecomment-1165427212

philippfromme:
We designed these icons a long time ago. I think they're ready for an overhaul.

nikku:
What we definitely want to do (mid-term) is to refresh our popup menu styles. You could follow up with it.

Reference for new pop-up menu styles should be the create/append anything popup: 

image

christian-konrad:
The latest design works for now in my opinion.
As there may be a growing set of actions for this menu(s), we can do an overhaul on all context menus or even the context pad itself in the future. Updating the icons could then also make sense (not only the alignment ones, but maybe all?) but for now, thanks for focusing on the functionality.
We should not keep 2 different context menu designs for an infinite time though.

nikku:
> We should not keep 2 different context menu designs for an infinite time though.

As we incorporate create/append anything into the core we'll get things aligned (again).

barmac:
I have two ideas how we can improve the popup menu so that it can be used for this PR and possible future improvements:

1. Minimal solution: Add grouping so that the buttons can be divided into align and distribute groups which can be seen on the UI sketch. In addition to that, set the popup menu name in the data attributes so that it can be used to style popup menu horizontally.

2. Powerful solution: Introduce a `popup-menu.render` event so that an extension can opt-in to render the popup menu on its own. As a fallback, the default rendering method is used. Thus, the extension developer has more flexibility if needed but can also use old APIs.

```javascript
eventBus.on('popup-menu.render', (menuId, node) => /* do stuff with node and return `true` if handled */)
```

For the current task of align/distribute menu, the minimal solution seems to be good enough. This is what I will implement but we can keep `2` in mind for the future.

nikku:
> Minimal solution: Add grouping so that the buttons can be divided into align and distribute groups which can be seen on the UI sketch. In addition to that, set the popup menu name in the data attributes so that it can be used to style popup menu horizontally.

+1 for doing that. If data-attributes are attached to the groups styling can be done naturally to achieve what you showcased.

barmac:
New challenge: Distribution and align actions can sometimes be cancelled internally when the action does not make sense (e.g. distribution of a task with boundary event). Notice how the popup menu does not close on the capture below for alignment. It closes for distribution actions, but only because I force it:

https://user-images.githubusercontent.com/28307541/175954902-db15f45f-321a-4fa7-a18d-f4807ed7459e.mov

The non-action buttons are similar to what Miro has, where the buttons are also displayed as disabled:

image

So I can see the following solutions:

1. Force close with no action (cf. distribution buttons on the capture)
2. Disable (close to alignment buttons on the capture)
3. Perform an initial check and don't display the buttons at all if action would be cancelled internally.

I'd rather not display disabled controls as a user cannot make any use of them. I am sure we should close the popup menu once a button is clicked. I think that in the end a solution based on rules (3) is the most intuitive. I doubt a user could be surprised there is no option to distribute elements if too few are selected. Still, in the initial iteration, we could have (1), and implement rules as a next step.

barmac:
I implemented option (1). Test this with:

`npx @bpmn-io/sr bpmn-io/bpmn-js#1680-add-alignment-and-distribution-buttons-to-context-pad -l bpmn-io/diagram-js#group-popup-menu-items`

barmac:
As diagram-js@8.6.0 is released, this is now ready for review.

barmac:
> As far as we never displayed disabled actions in the past in our context pad(s), we should go for 3 in the future IF we try to implement such rules; but we will first validate the demand for this.

I've created an issue for this: https://github.com/bpmn-io/bpmn-js/issues/1691

> @barmac after merge, can you reach out, so we can implement events for tracking opening the alignment action menu and actually applying an alignment?

OK so metrics first, then the rules.

barmac:
Thanks @nikku, I passed it to the epic issue: https://github.com/camunda/product-hub/issues/61#issuecomment-1169802249

barmac:
As this PR will include rules implementation which is WIP now, I'm changing it to draft again.

philippfromme:
I don't think being too strict makes for a good experience. In the spirit of our tools we should always let the user use the alignment tools if more than one element is selected. Behind the scenes we should figure out a reasonable result for the given selection.

As I said before, I'd also want to see the icons refreshed.

barmac:
Thank you both for feedback.

> Selection of align/distributable elements not intuitive to me. I'd assume that I can select a sub-process and its children along with other elements and still align (ignoring the children). I#d expect selecting sub-process children + sub process sibling and be able to align (why not)?

Which elements should move given I align middle the selection below?

image

> Icons are to big, they should have a little bit of spacing 1px (?) around them

Do you mean "too big", as they should be smaller (say 20 or 18px instead of 22px)? I get that spacing is too big for you right now, correct?

> Learning from Miro you don#t need three elements to visualize alignment direction => Less noise on the diagram.

I'll see what I can do but I'd rather prioritize the correct behavior.

> I don't think being too strict makes for a good experience. In the spirit of our tools we should always let the user use the alignment tools if more than one element is selected. Behind the scenes we should figure out a reasonable result for the given selection.

I understand that the tool behavior should be more lenient. The answer to the top question I asked would be that it should align the subprocess and the task only:

image

barmac:
OK so now you can test how it works if we hide buttons only when we know upfront there is nothing to align or distribute:

`npx @bpmn-io/sr bpmn-io/bpmn-js#1680-add-alignment-and-distribution-buttons-to-context-pad-strict -l bpmn-io/diagram-js#remove-built-in-distribution-filter`

barmac:
+ smaller icons / less space in between

barmac:
When I select both parent and children, the distribution is broken though 😅 

https://user-images.githubusercontent.com/28307541/176672246-416f5ae9-7faf-42dc-896b-1aba1e653cd9.mov

nikku:
> Which elements should move given I align middle the selection below?

The two parent elements (task + sub-process). In general, consider all "shared parents" for the align operation.

> Do you mean "too big", as they should be smaller (say 20 or 18px instead of 22px)? I get that spacing is too big for you right now, correct?

The icon is too big, exactly. Spacing is fine.

barmac:
This is how the context pad icon would look like if we make it 20px x 20px and add a 1px padding:

image

barmac:
> The two parent elements (task + sub-process). In general, consider all "shared parents" for the align operation.

If I get it correctly, the middle alignment of selection below...

image

...will effect in this:

image

barmac:
Let's try this approach out. I think it looks pretty neat and the implementation is super brief due to existing diagram-js helper ^^

barmac:
https://user-images.githubusercontent.com/28307541/176687997-9a409746-5fb4-49ca-a2e4-956adcb52846.mov

nikku:
@barmac Fantastic. Exactly as I'd expect it.

barmac:
Some experiments with the general icon:

image
image

image

christian-konrad:
Wow, now this feels mature. Please let's do this!
On the icons, I like the first and second style (slightly rounded); as the others said, the old ones add too much noise; we should take the time for this adjustment. Either 1 or 2 depending on how it looks like when all icons are in place (maybe 1 is better due to the smaller inner margin in the icon, which makes it look more cohesive once all icons are displayed)

nikku:
Does this also address https://github.com/bpmn-io/bpmn-js/issues/1681?

barmac:
> Does this also address #1681?

Indeed, with the new most recent changes it fixes the problem.

barmac:
> Wow, now this feels mature. Please let's do this! On the icons, I like the first and second style (slightly rounded); as the others said, the old ones add too much noise; we should take the time for this adjustment. Either 1 or 2 depending on how it looks like when all icons are in place (maybe 1 is better due to the smaller inner margin in the icon, which makes it look more cohesive once all icons are displayed)

Thanks for your feedback. I will prepare then the remaining icons based on 1.

barmac:
IMO the space between elements seems to be too small. They appear a bit blurry to me:

image

barmac:
New attempt: 
image

barmac:
A complete set:

image

philippfromme:
> A complete set:
> 
> image

I think this is starting to look good. Our icons don't feature grey, though, so this would be a change.

nikku:
Black and white is a relict of the past (being a font). We can add one more color, if that benefits our UX. :slightly_smiling_face:

barmac:
This is now ready for review.

barmac:
Screen Shot 2022-07-06 at 11 57 54
Just added some changes so that it looks OK on IE11. It won't be 1:1 as in the modern browsers but we seem to not care anymore since we use css variables 🤡

marstamm:
I played around with it a bit and noticed something strange. On the simple process, if I align everything `center`, then `top`, the flows get reconnected:

Recording 2022-07-06 at 14 21 41

Obviously this is not what this tool was intended for, but would be interesting to know what causes a reconnect in the first place. Is `dropOnFlow` triggered during alignment?

philippfromme:
> Is `dropOnFlow` triggered during alignment?

That is what's happening, yes. We might have to pass a hint preventing that behavior.

barmac:
> I would say it is an edge case that we don't have to fix right now

I've created an issue for this: https://github.com/bpmn-io/bpmn-js/issues/1693

philippfromme:
Not sure why the distribution icons don't show up when selecting 3 participants. Is that expected? Couldn't find anything in the code.

image

philippfromme:
Otherwise, this is looking really good and very solid. Looking forward to getting that thing out of the door. 💪🏻

barmac:
OK it doesn't make sense. For some unknown to me reason, the legacy code disallowed distribution of participants.

barmac:
I will add a test case for this.

barmac:
OK done @philippfromme

barmac:
This broke other tests 🙈

barmac:
OK Fixed.

philippfromme (approved):
Great job! 🚀

PR #1685, opened 2022-06-24
test: add generic logging facility

nikku:
This allows us to enable different sorts of logging conveniently
via the test helper and control it via query parameters.

-----

The basic idea is that you can control via a query parameter if you'd like to log what is happening XML wise during a test run.


pinussilvestrus (approved):
Less configuration overhead, I like 👍 

http://localhost:9877/debug.html?log=save-xml

image

PR #1694, opened 2022-07-07
fix: properly size icons in distribute/align

barmac:

### DIFF
--- a/assets/bpmn-js.css
+++ b/assets/bpmn-js.css
@@ -129,15 +129,14 @@
 }
 
 [data-popup="align-elements"] .djs-popup-body .entry {
-  height: 20px;
-  width: 20px;
-
   padding: 6px 8px;
 }
 
 [data-popup="align-elements"] .djs-popup-body .entry img {
-  height: 100%;
-  width: 100%;
+  display: block;
+
+  height: 20px;
+  width: 20px;
 }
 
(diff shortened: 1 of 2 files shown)


marstamm (approved):
Works on demo page 🚀

assets/bpmn-js.css
```diff
@@ -129,15 +129,14 @@
...
-
   padding: 6px 8px;
 }
 
 [data-popup="align-elements"] .djs-popup-body .entry img {
-  height: 100%;
-  width: 100%;
+  display: block;
+
+  height: 20px;
```
nikku, line 138: Is there no other way than hardcoding (to the parent height)?

For me this looks like duplicate magic numbering. We already define the entry height in diagram-js (cf. https://github.com/bpmn-io/diagram-js/blob/develop/assets/diagram-js.css#L512).

barmac: Hmm, perhaps we could just use `1em` here. We need to set the image size explicitly because otherwise it shrinks.

barmac: But then the em value is also a magic number as it uses the line-height / font-size ratio 🙈

nikku: Isn't the issue that `width: 100%` does not mean a thing? I#ll have a look.

PR #1711, opened 2022-08-13
fix(modeling): do not alter attrs passed to `ElementFactory#create`

nikku:
Ensures we do not alter input parameters. Good practice.


philippfromme (approved):
Fixed a typo. Good job! :rocket

PR #1718, opened 2022-08-17
Bump dependencies

nikku:
Basic housekeeping on our dependencies.


nikku:
As it turns out we

* cannot upgrade to Webpack@5 due to it generating `() => { }` bundles
* cannot upgrade test dependencies as one secretly (no major version bump) introduced ES6 syntax

nikku:
All green and ready to be merged at last!

barmac (approved):

PR #1722, opened 2022-08-19
Remove PhantomJS from test suite

nikku:
As discussed within the team it is time to drop some more legacy from our code base.

We do target decently modern browsers at this time, dropped IE support (no active QA against it).

Time to also drop PhantomJS for the sake of better tests, implementation, and future :rocket:


test/config/karma.unit.js
```diff
@@ -7,7 +7,7 @@ var singleStart = process.env.SINGLE_START;
 var coverage = process.env.COVERAGE;
 
 // configures browsers to run test against
-// any of [ 'ChromeHeadless', 'Chrome', 'Firefox', 'IE', 'PhantomJS' ]
+// any of [ 'ChromeHeadless', 'Chrome', 'Firefox' ]
```
marstamm, line 10: As we now test against Safari, we could add it here as well

```suggestion
// any of [ 'ChromeHeadless', 'Chrome', 'Firefox', 'Safari' ]
```

marstamm (approved):

test/config/karma.unit.js
nikku: Already mentioning in the [contributing guide](https://github.com/bpmn-io/bpmn-js/commit/5680826e9e272da3713811e738738b3a14afe479).

PR #1723, opened 2022-08-19
Update to webpack@5

nikku:
+ drop raw-loader (replaced by webpack built-ins)

-----

Depends on https://github.com/bpmn-io/bpmn-js/pull/1722


marstamm (approved):
Nice one. I overlooked this one on the modeler update: 
https://github.com/camunda/camunda-modeler/blob/develop/client/webpack.config.js#L82

I'll remove raw-loader amd file-loader from the modeler as well 👍

nikku:
Could you also have a look into https://github.com/bpmn-io/bpmn-js/pull/1722? Then we could merge both :).

marstamm:
Will do

PR #1727, opened 2022-08-22
deps: update to `bpmn-moddle@7.1.3`

barmac:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2324,13 +2324,13 @@
       "dev": true
     },
     "bpmn-moddle": {
-      "version": "7.1.2",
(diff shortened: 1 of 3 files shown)


barmac:
I don't know where the `const` failure comes from. Anyway, if we are releasing this with `bpmn-js@9.4.0`, I could as well target this at `develop`.

barmac:
I missed https://github.com/bpmn-io/bpmn-js/pull/1726#issuecomment-1222147263, sorry.

nikku:
[This](https://github.com/bpmn-io/bpmn-js/pull/1727/commits/5e554bae5bfcab64156a1dca2194aa069160dbc2) is where it is coming from.

barmac:
Thanks Nico!

PR #1788, opened 2022-12-08
deps: update to diagram-js@11.3.0

smbea:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "bpmn-moddle": "^8.0.0",
         "css.escape": "^1.5.1",
-        "diagram-js": "^11.2.1",
+        "diagram-js": "^11.3.0",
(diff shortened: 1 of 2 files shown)


barmac:
Any idea why it still displays all of the entries with up to 3 dashes?

image

barmac:
Anyway, it fixes the issue with "replace" matching anything.

barmac:
OK I think I know :)

https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/popup-menu/PopupMenuComponent.js#L98

```javascript
  useEffect(() => {
    if (!searchable) {
      return;
    }

    const filter = entry => {
      if (!value) {
        return true;
      }

      const search = [
        entry.description || '',
        entry.label || ''
      ]
        .join('---')
        .toLowerCase();

      return value
        .toLowerCase()
        .split(/\s/g)
        .every(term => search.includes(term));
    };
```

barmac (approved):

PR #1807, opened 2023-01-18
fix(element-factory): add missing attributes

smbea:
Came up in the context of https://github.com/bpmn-io/bpmn-js/pull/1802

* Event subprocesses were being created as normal subprocesses
* Non interrupting boundary events were being created as interrupting events


nikku (approved):
Awesome :+1:

PR #1809, opened 2023-01-18
Append keyboard shortcut

### LINKED ISSUE #1808: Add keyboard shortcuts for append
### What should we do?

Add a shortcut for opening the replace menu ('a', 'A')

### Why should we do it?

As part of the replace anything epic, see https://github.com/bpmn-io/bpmn-js/issues/1627

smbea:
Closes #1808


nikku (approved):
Works for me.

PR #1811, opened 2023-01-19
Add "create element" to palette

### LINKED ISSUE #1798: Add create to palette
__What should we do?__

Add a "create anything" entry to the palette which provides a menu with all possible elements.

What it should look like (but with bpmn elements):
Screenshot 2023-01-10 at 16 38 45

 __To do:__
- [ ] Add entry to palette 
- [ ] Add create menu entries provider

__Why should we do it?__

To support create anything, see https://github.com/bpmn-io/bpmn-js/issues/1627

smbea:
Closes #1798


lib/features/palette/PaletteProvider.js
```diff
@@ -32,7 +36,9 @@ PaletteProvider.$inject = [
   'lassoTool',
   'handTool',
   'globalConnect',
-  'translate'
+  'translate',
+  'popupMenu',
```
nikku, line 40: One general question that I had is "should we make create/append anything" available as a pluggable module => You can either have it or not.

This question comes up any time I see us hooking into existing behaviors, where we could just build the feature self-contained.

Why could that be beneficial: Because users then have the choice to benefit from our modular architecture, either use it, or not.

(Same could be said for other things of course, such as aligmnent utilities, ...)

lib/features/popup-menu/CreateMenuProvider.js
```diff
@@ -0,0 +1,111 @@
...
+      target,
+      description,
+      group,
+      search
+    } = option;
+
+    const targetAction = this._createEntryAction(target);
+
+    entries[`create-${actionName}`] = {
+      label,
```
nikku, line 64: We should ensure we translate this label + the group label (cf. https://github.com/bpmn-io/bpmn-js/pull/1810). As of today you translate within the provider, not inside the menu.

lib/icons/Icons.js
```diff
@@ -3,7 +3,9 @@
  * and then replace respective icons with the optimized data URIs in `./dist`.
  */
 const appendIcon = 'data:image/svg+xml,%3Csvg%20width%3D%2222%22%20height%3D%2222%22%20viewBox%3D%220%200%205.82%205.82%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%3E%0A%20%20%3Cpath%20d%3D%22M1.3%203.4c.3%200%20.5-.2.5-.5s-.2-.4-.5-.4c-.2%200-.4.1-.4.4%200%20.3.2.5.4.5zM3%203.4c.2%200%20.4-.2.4-.5s-.2-.4-.4-.4c-.3%200-.5.1-.5.4%200%20.3.2.5.5.5zM4.6%203.4c.2%200%20.4-.2.4-.5s-.2-.4-.4-.4c-.3%200-.5.1-.5.4%200%20.3.2.5.5.5z%22%2F%3E%0A%3C%2Fsvg%3E';
+const createIcon = 'data:image/svg+xml,%3Csvg%20width%3D%2246%22%20height%3D%2246%22%20viewBox%3D%22-2%20-2%209.82%209.82%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%3E%0A%3Cpath%20d%3D%22M1.3%203.4c.3%200%20.5-.2.5-.5s-.2-.4-.5-.4c-.2%200-.4.1-.4.4%200%20.3.2.5.4.5zM3%203.4c.2%200%20.4-.2.4-.5s-.2-.4-.4-.4c-.3%200-.5.1-.5.4%200%20.3.2.5.5.5zM4.6%203.4c.2%200%20.4-.2.4-.5s-.2-.4-.4-.4c-.3%200-.5.1-.5.4%200%20.3.2.5.5.5z%22%2F%3E%0A%3C%2Fsvg%3E';
```
nikku, line 6: Isn't it exactly the same icon? In this case we could simply `const creatIcon = appendIcon`.

nikku (commented):
Works like a charm.

What I'm missing is a keyboard shortcut, and semantic ordering. I guess we'll follow up on these.

Added a bunch of comments for nit-picks.

lib/features/palette/PaletteProvider.js
nikku: > [not include these in the base modeler](https://github.com/bpmn-io/bpmn-js/blob/11a7666ff111d84c9d19f400c71cf4fed54e779b/lib/Modeler.js#L180)

We can include it in the base modeler, as there exists the ["really base modeler"](https://github.com/bpmn-io/bpmn-js/blob/11a7666ff111d84c9d19f400c71cf4fed54e779b/lib/BaseModeler.js#L180) people can use to customize (pick only what they actually need).

lib/icons/Icons.js
nikku: Happy to keep both icons for now then.

lib/features/palette/PaletteProvider.js
nikku: Consider combining both in a shared place to simplify things. "create/append anything" is a feature.

barmac:
So the test suite fails on CI but for Mac only. On my Mac, it works fine though 🤔

lib/Modeler.js
```diff
@@ -20,6 +20,7 @@ import ConnectionPreviewModule from 'diagram-js/lib/features/connection-preview'
 import ContextPadModule from './features/context-pad';
 import CopyPasteModule from './features/copy-paste';
 import CreateModule from 'diagram-js/lib/features/create';
+import CreateAppendElements from './features/create-append-elements';
```
nikku, line 23:
```suggestion
import CreateAppendAnythingModule from './features/create-append-anything';
```

lib/Modeler.js
```diff
@@ -190,6 +191,7 @@ Modeler.prototype._modelingModules = [
   ContextPadModule,
   CopyPasteModule,
   CreateModule,
+  CreateAppendElements,
```
nikku, line 194: Maybe rename to `CreateAppendAnything`?

lib/features/create-append-elements/CreateMenuProvider.js
```diff
@@ -0,0 +1,113 @@
+import { CREATE_OPTIONS } from '../popup-menu/util/OptionsUtil';
```
nikku, line 1: Can likely be pulled into this module, as it is only used by CreateAppendAnything as it looks like?

nikku:
Very nice improvement moving stuff to a dedicated component, I think.

barmac:
Found a way to break it:

https://user-images.githubusercontent.com/28307541/214008728-9347dc1e-34b6-4ad0-8422-794fc74457e1.mov

Easy way to fix this is to disable this functionality for all connections.

barmac:
OK done: https://github.com/bpmn-io/bpmn-js/issues/1813

barmac:
Sorry misclick

lib/features/create-append-anything/CreateMenuProvider.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+export default function CreateMenuProvider(
+    elementFactory, popupMenu, create,
+    autoPlace, mouse, translate
+) {
+  this._elementFactory = elementFactory;
+  this._popupMenu = popupMenu;
+  this._create = create;
+  this._autoPlace = autoPlace;
+  this._create = create;
```
barmac, line 14:
```suggestion
  this._create = create;
  this._autoPlace = autoPlace;
```

lib/features/create-append-anything/CreatePaletteProvider.js
```diff
@@ -0,0 +1,61 @@
...
+      title: translate('Create element'),
+      action: {
+        click: function(event) {
+          const cursor = event && { x: event.x, y: event.y };
+
+          const position = event && {
+            x: cursor.x + 35,
+            y: cursor.y + 10,
+            cursor
+          };
```
barmac, line 46: Not sure if we want this behavior:

https://user-images.githubusercontent.com/28307541/214012932-6922d29e-96f7-4311-ae74-f07185e3ebc2.mov

Still it's ok to leave it as is if we can't get the entry position easily.

barmac (commented):
LGTM. Check out the remaining comment and merge when you think it's ready. No need for another round of review from my perspective 🚀

nikku:
:tada:

Ken-Scofield:
Why missing "Create" in Palette? or how to add it
no-create

PR #1815, opened 2023-01-23
Move append to `create-append-anything` module

smbea:
This is a follow up from https://github.com/bpmn-io/bpmn-js/pull/1811#discussion_r1082321170

Moving this causes what seems like a big PR so let me know if I can help make the review easier in any way.

What I suggest is to check out commits https://github.com/bpmn-io/bpmn-js/commit/9f302000 and https://github.com/bpmn-io/bpmn-js/commit/d5ad82ba which add something new (encapsulating editor actions/keyboard shortcuts). The rest is pretty much the same but on a different location (but of course feel free to check it out)


lib/features/create-append-anything/index.js
```diff
@@ -1,20 +1,38 @@
...
+import AppendContextPadProvider from './AppendContextPadProvider';
 
 export default {
   __depends__: [
     PaletteModule,
     PopupMenuModule,
-    AutoPlace
+    AutoPlaceModule,
+    ContextPadModule,
+    EditorActionsModule,
```
nikku, line 21: We should not be dependent on this, but make it an optional dependency.

lib/features/create-append-anything/index.js
```diff
@@ -1,20 +1,38 @@
...
+    KeyboardModule
```
nikku, line 22: We should not be dependent on this, but make it an optional dependency.

lib/features/create-append-anything/CreateAppendEditorActions.js
```diff
@@ -0,0 +1,44 @@
+import { assign } from 'min-dash';
+
+/**
+ * Registers and executes BPMN specific editor actions.
+ *
+ * @param {Injector} injector
+ */
+export default function CreateAppendEditorActions(injector, editorActions) {
```
nikku, line 8: Consider to make this optionally depend on `editorActions`.

You'll otherwise not be able to properly compose it.

Alternative (but maybe overkill):

* add `create-append-anything/editor-actions` and `keyboard` modules

nikku (commented):
One small topic you may want to follow up on to improve composability:

* Do not hard wire keyboard and editor actions as dependencies, as the modeler works without those usually.

Other than that, great! A really good showcase how this improves maintainability to have such feature contained in a modular fashion.

nikku (approved):
Wonderful.

PR #1818, opened 2023-01-25
chore(create-append-anything): move rules to dedicated module

smbea:
Continuation of https://github.com/bpmn-io/bpmn-js/pull/1815
We missed moving the append anything rules


nikku (approved):

nikku:
Nice one!

PR #1822, opened 2023-01-30
feat: use single arrow for connection icon

barmac:
image

Related to https://github.com/camunda/camunda-modeler/issues/2282

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
         "@rollup/plugin-terser": "^0.1.0",
         "babel-loader": "^9.1.0",
         "babel-plugin-istanbul": "^6.1.1",
-        "bpmn-font": "^0.11.0",
+        "bpmn-font": "^0.12.0",
         "camunda-bpmn-moddle": "^4.0.1",
         "chai": "4.1.2",
         "chai-match": "^1.1.1",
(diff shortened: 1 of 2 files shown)


nikku (changes requested):
To reduce the visual disturbance caused by this change I'd suggest to use a connection icon that is "in line" with other connections in the context pad:

image

barmac:
I can do that but it will require changes in bpmn-font. 

BTW note that the default flow angle is the same as the icon I used here:

image
image

nikku:
Asked for quick validation via with a bunch of other folks. If it is just me, let's keep it as you propose.

barmac:
Hmm I think instead of bpmn-font changes it can be implemented in bpmn-js via icons module.

barmac:
image

I'm not happy with this result. It looks too crispy, and even setting `shape-rendering` doesn't help.

nikku:
@barmac We can make minor improvements to the icon to make it look more crisp:

image.

However we'll not be able to fix the issue that the new icon has a weaker visual presence then the old one. This is simply because it takes less space.

barmac:
I experimented with text annotation:

image

barmac:
The wrench:
image

nikku:
@barmac Let's put up a demo where we can see this end-to-end.

barmac:
I prepared a bpmn-font update and will post a demo link soon: https://github.com/bpmn-io/bpmn-font/pull/19

barmac:
I've changed the PR to use the font from the linked pull request but now the icons appear broken. Not sure if this is due to some changes at fontello or I did sth wrong:

image

nikku:
@barmac This is due to

image

barmac:
OK so it's fontello's fault. I will adjust the PR.

barmac:
This is how it looks like with adjustments made:

image

marstamm:
I love how the lines match up now, rather than just being parallel to each other ❤️ 
image (2)

nikku:
I still love the old connect icon in the way it fills a square place, but hey, let's embrace the future.

Maybe a future follow-up could be to make the connection line width the same width like the annotation line (text annotation).

philippfromme:
I think it looks just fine.

nikku:
@philippfromme Consider https://github.com/bpmn-io/bpmn-js/pull/1825 as an alternative. Which one works better?

| This PR | https://github.com/bpmn-io/bpmn-js/pull/1825 |
| :--- | :--- |
| image | image |

philippfromme:
@nikku The second one works much better.

marstamm:
+1 for #1825

nikku:
Incorporated the final icons via https://github.com/bpmn-io/bpmn-js/pull/1822/commits/2e6b976dfa7e52285150b27f4f10fc05a315cb3c:

image

nikku (approved):

PR #1827, opened 2023-02-01
Improve shape rendering / join path `round`

nikku:
This PR ensures that we render paths on the diagram with `round` linecap / linejoin. This leads to a (slightly) improved, less harsh for the eye visual representation for shapes, connections, and markers.

## Before 

screenshot vdGOq3

## After

screenshot ALOcmc

---

Noteworthy glitches fixed:

* Sequence Flow end marker has _one_ out of three sharp corners
* Data Object looks aweful

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -106,8 +106,7 @@ export default function BpmnRenderer(
     var {
       ref = { x: 0, y: 0 },
       scale = 1,
-      element,
-      attrs
+      element
     } = options;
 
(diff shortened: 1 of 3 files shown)


nikku:
Test cases are [currently failing](https://github.com/bpmn-io/bpmn-js/actions/runs/4064211285/jobs/6997372865). Happy to clean that up if we want to pursue this direction.

barmac (approved):

barmac:
I believe this is the right direction.

PR #1832, opened 2023-02-03
chore(CHANGELOG): update to 11.2.0

smbea:
It's a big one!

Note: I added all the changes for append/create in the same point because they are new features completely. But let me know if you think they should be explicitly itemized


nikku (approved):

nikku:
Good stuff!

Issue #1621, opened 2022-03-22
Improve semantic ordering in replace / append menu for "implementation"


nikku:
__Is your feature request related to a problem? Please describe.__

As a tool we want our users to [get work done as efficiently as possible](https://github.com/bpmn-io/design-principles#do-the-hard-work-to-make-it-simple). Upgrading tasks via the replace menu to make them executable is an important part of the modeling interaction.

To date, however, we don't provide clear guidance to the user on what is possible / relevant. In fact, considering the options provided by the change / replace menu, you can see that the order is historically grown / accidental:

![image](https://user-images.githubusercontent.com/58601/159464788-31686246-bc56-4a4e-9921-672ca3d724e6.png)

__Describe the solution you'd like__

* [x] Improve ordering of choices provided to our users so that they are guided (use common, well known elements over esotherics)
* [x] Cater towards different use-cases (make ordering pluggable)? (https://github.com/bpmn-io/bpmn-js/issues/1621#issuecomment-1382139235)

__Describe alternatives you've considered__

Remove discouraged options (i.e. "manual task") would be a "execution only" option. 

Do not have a fixed order, but also incorporate "recently used" into the mix (cf. [know the context](https://github.com/bpmn-io/design-principles#know-the-context)).

__Additional context__

Getting this right contributes to "speed up time to implemented" (https://github.com/camunda/web-modeler/issues/860).

We shall consider this in the context of the [new change / append UIs](https://github.com/bpmn-io/connectors-modeling-demo) shipping with the web modeler. There we can rely on search for _rare_ items to be discoverable.

----

This is addressed through https://github.com/bpmn-io/bpmn-js/issues/1627.

nikku:
Moving to backlog as we cannot immediately act upon this; needs further discovery.

nikku:
Ordering is already pluggable today:

* A user can define an extension and re-order the elements at will.

nikku:
Feedback shared via https://github.com/camunda/product-gaps/issues/73:

As a user I'd like to replace from one event type to another without loosing the event definition contents (i.e. message configured).

PR #1836, opened 2023-02-06
Improve semantic ordering / filter rare elements

nikku:
This closes https://github.com/bpmn-io/bpmn-js/issues/1621 in two ways:

* Initially rare elements are not shown in create/append anything menus
* User and service tasks are moved up in the menu to make them more prominent

---

capture Tv7P1h_optimized


philippfromme (approved):
Fixed a typo in the changelog. Cool stuff! 🚀

PR #1843, opened 2023-02-10
fix(append-anything): element connection via dragstart

smbea:
Related to https://github.com/camunda/camunda-modeler/issues/3436


nikku (approved):
Great stuff.

PR #1855, opened 2023-02-24
feat: add root elements to definitions for updates

barmac:
This is so that I don't need to manually add root element when updating messageRef or event definitions.

### DIFF
--- a/test/spec/features/modeling/behavior/RootElementReferenceBehaviorSpec.js
+++ b/test/spec/features/modeling/behavior/RootElementReferenceBehaviorSpec.js
@@ -189,6 +189,204 @@ describe('features/modeling - root element reference behavior', function() {
 
       });
 
+      describe(`${type} (modeling#updateProperties)`, function() {
+
+        var boundaryEvent,
+            rootElement;
+
+        describe('should add root element if not added to diagram', function() {
+
+          beforeEach(inject(function(bpmnFactory, elementRegistry, modeling) {
+
+            // given
+            boundaryEvent = elementRegistry.get('BoundaryEvent');
+            rootElement = bpmnFactory.create(`bpmn:${capitalizeFirstChar(type)}`);
+            var eventDefinition = bpmnFactory.create(`bpmn:${capitalizeFirstChar(type)}EventDefinition`, {
+              [`${type}Ref`]: rootElement
+            });
+
+            // when
+            modeling.updateProperties(boundaryEvent, {
+              eventDefinitions: [
+                eventDefinition
+              ]
+            });
+          }));
+
+
+          it('<do>', function() {
+
+            // then
+            expect(hasRootElement(rootElement)).to.be.true;
+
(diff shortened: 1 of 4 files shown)


test/spec/features/modeling/behavior/RootElementReferenceBehaviorSpec.js
```diff
@@ -316,6 +508,163 @@ describe('features/modeling - root element reference behavior', function() {
...
+            // when
+            commandStack.redo();
+
+            // then
+            expect(hasRootElement(rootElement)).to.be.true;
+          }));
+        });
+
+
+        it('should NOT add on modeling#updateProperties', inject(function(
```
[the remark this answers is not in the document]
barmac: Thanks for the feedback. I will adjust the tests.

test/spec/features/modeling/behavior/RootElementReferenceBehaviorSpec.js
```diff
@@ -316,6 +510,163 @@ describe('features/modeling - root element reference behavior', function() {
...
+
+
+        it('should NOT add on modeling#updateProperties if already added', inject(function(
+            bpmnFactory, bpmnjs, elementRegistry, modeling
+        ) {
+
+          // given
+          task = elementRegistry.get(id);
+
+          var message = bpmnFactory.create('bpmn:Message', { id: 'Message_2' });
```
nikku, line 629: I assume `bpmnFactory#create` triggers a hidden side-effect and adds the message?

barmac: Not really, the message is added as a reference via https://github.com/bpmn-io/bpmn-js/pull/1855/files#diff-eb9635ee33330272486aec3b24a14c913f6d10c8903f3fdc29b3519908adb005R637, and then the tested behavior reacts on `element.updateProperties` and adds the Message to the definitions#rootElements.

nikku: So the description should read `should add only once`, right?

barmac:
It seems there is a bug in the implementation. Namely, when I add a Message with ID of already existing Message via `modeling#updateProperties`, the ref should be replaced but it's not.

test/spec/features/modeling/behavior/RootElementReferenceBehaviorSpec.js
barmac: Hmm it should not add root element if it is already present on the diagram. However, the test case is broken as we just duplicate ID instead of the exact reference. I need to think about it.

barmac: I will update the test cases so that they clearly show what we want to support.

barmac:
OK I think it should be now clear.

nikku (approved):

PR #1870, opened 2023-03-17
Major rollup upgrade to v3

smbea:
* Updates rollup and related plugins
* Makes necessary changes to rollup config


rollup.config.mjs
```diff
@@ -81,6 +83,7 @@ function banner(bundleName, minified) {
       : 'banner'
   );
 
+  const __dirname = path.resolve();
```
nikku, line 86: Was not aware that this is how it works. I always did [this](https://stackoverflow.com/a/50052194).

nikku: If it works, great :)

nikku (approved):

PR #1881, opened 2023-04-04
wip: Add type declarations to Modeling.js

miyuesc:
1. Add type declarations to Modeling.js
2. The options parameter in the saveXML method is optional


nikku:
@miyuesc We're currently in the process of generating type definitions for bpmn-js, cf. https://github.com/bpmn-io/bpmn-js/pull/1866. Does your PR work on top of these changes?

miyuesc:
I'm sorry, I didn't notice it was used this way before.

Issue #1872, opened 2023-03-21
`Modeling#updateModdleProperties` does not add label for events name modification


barmac:
### Describe the Bug

When I call `modeling.updateModdleProperties` in order to change element's `name`, the label is not created.

### Steps to Reproduce

https://codesandbox.io/s/relaxed-kilby-4bh0xi?file=/src/index.js

### Expected Behavior

Label is added as with `modeling.updateProperties`.

### Environment

 - Browser: [e.g. IE 11, Chrome 69]
 - OS: [e.g. Windows 7]
 - Library version: [e.g. 2.0.0]

barmac:
One more bug related to this: We don't remove BPMNLabel when name is erased via `modeling.updateProperties` in external label elements.

barmac:
Architecture sketch:

![image](https://user-images.githubusercontent.com/28307541/233029190-54e9580a-41a7-49f0-9da4-21594699a044.png)

barmac:
Fixed via https://github.com/bpmn-io/bpmn-js/pull/1894

Issue #1893, opened 2023-04-19
BPMNLabel is not removed when name is erased on an external label element


barmac:
One more bug related to this: We don't remove BPMNLabel when name is erased via `modeling.updateProperties` in external label elements.

_Originally posted by @barmac in https://github.com/bpmn-io/bpmn-js/issues/1872#issuecomment-1511101056_

barmac:
Hi,

We need a test case which verifies that BPMNLabel is removed when an external label is removed, and the implementation which does it. I envision this to be implemented in the BpmnUpdater, specifically in this function: https://github.com/bpmn-io/bpmn-js/blob/1872-fix/lib/features/modeling/BpmnUpdater.js#L300 Notice that the link refers to a PR which will be probably merged soon.

VISHESH0932:
Can i work on this issue?

nikku:
@VISHESH0932 [pull requests](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests) are open, happy to take a contribution.

PR #1894, opened 2023-04-19
Update label on modeling.updateModdleProperties

barmac:
Closes https://github.com/bpmn-io/bpmn-js/issues/1872

I extracted one of the bugs into a separate issue: https://github.com/bpmn-io/bpmn-js/issues/1893


lib/features/label-editing/cmd/UpdateLabelHandler.js
```diff
@@ -77,7 +53,7 @@ export default function UpdateLabelHandler(modeling, textRenderer, bpmnFactory)
 
     setLabel(label, text, labelTarget !== label);
 
-    ensureInternalLabelDi(element, text);
+    // ensureInternalLabelDi(element, text);
```
nikku, line 56: We'd want to get rid of this line, right?

lib/features/modeling/BpmnUpdater.js
```diff
@@ -18,7 +18,11 @@ import {
 
 import { isAny } from './util/ModelingUtil';
 
-import { isLabel } from '../../util/LabelUtil';
+import {
+  isLabel,
+  isLabelExternal
+} from '../../util/LabelUtil';
```
nikku, line 24: Combine both imports.

lib/features/modeling/behavior/LabelBehavior.js
```diff
@@ -86,7 +79,16 @@ export default function LabelBehavior(
   CommandInterceptor.call(this, eventBus);
 
   // update label if name property was updated
-  this.postExecute('element.updateProperties', function(e) {
+  this.postExecute('element.updateProperties', onPropertyUpdate);
```
nikku, line 82: Do we also cover text annotation and groups (category) here? Is that needed?

nikku (changes requested):
Nice one! Please checkout my remarks.

lib/features/label-editing/cmd/UpdateLabelHandler.js
barmac: Yeah, my bad

lib/features/modeling/BpmnUpdater.js
barmac: I moved LabelUtil up while maintaining backwards compatibility: https://github.com/bpmn-io/bpmn-js/pull/1894/commits/b05649220b2206956dd8acd113f01d7f3de472f0

lib/features/modeling/behavior/LabelBehavior.js
barmac: As you can see below, TextAnnotation is still covered. Not sure about the groups. However, I don't want to creep the scope of this PR.

lib/features/modeling/behavior/LabelBehavior.js
```diff
@@ -133,20 +131,7 @@ export default function LabelBehavior(
       return;
     }
 
-    var labelCenter = getExternalLabelMid(element);
-
-    // we don't care about x and y
-    var labelDimensions = textRenderer.getExternalLabelBounds(
-      DEFAULT_LABEL_DIMENSIONS,
```
nikku, line 140: Where did the `DEFAULT_LABEL_DIMENSIONS` go? We don't need it anymore?

nikku (approved):
Good improvement and refactoring :tada:

PR #1896, opened 2023-04-20
fix: fix remaining JSDocs

miyuesc:


nikku:
Thanks for your contribution. Great stuff :tada:.

I've fixed one issue that caused our type generation to fail, but otherwise this looks fabulous.

miyuesc:
Many front-end developers in China are currently using this library, and it's my pleasure to provide some help for this.

nikku:
Released with `v13.0.1`: https://github.com/bpmn-io/bpmn-js/blob/develop/CHANGELOG.md#1301.

PR #1898, opened 2023-04-21
fix: correct various JSDocs

miyuesc:


nikku:
Thanks for this contribution.

PR #1917, opened 2023-05-23
Allow event rendering without icons

nikku:
This adds a new option to the `BpmnRenderer` to render events without icons. The use-case is customizing (cf. https://github.com/bpmn-io/element-template-icon-renderer/issues/6).

For detailed usage, checkout https://github.com/bpmn-io/bpmn-js/pull/1917/files#diff-c8205dfe6f35f2b861d97b83d680753e34aea03b27bbeee25bc42d8579604807R52.

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -444,8 +444,8 @@ export default function BpmnRenderer(
   }
 
   function as(type) {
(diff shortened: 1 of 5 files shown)


philippfromme (approved):
Nice and simple. I like! ✅ Use at your own risk. 💣

PR #1920, opened 2023-06-05
Load icons as html

### LINKED ISSUE #1918: Follow html pattern for icon loading
### What should we do?

Replace instances where we use "embedded icons" and replace it with html, like https://github.com/bpmn-io/bpmn-js-create-append-anything/pull/7

### Why should we do it?

To establish a single pattern and be consistent across our stack

smbea:
Close #1918

### DIFF
--- a/lib/features/align-elements/AlignElementsContextPadProvider.js
+++ b/lib/features/align-elements/AlignElementsContextPadProvider.js
@@ -70,7 +70,7 @@ AlignElementsContextPadProvider.prototype._getEntries = function() {
     'align-elements': {
       group: 'align-elements',
       title: self._translate('Align elements'),
-      imageUrl: ICONS['align'],
+      html: `<div class="entry">${ICONS['align']}</div>`,
       action: {
         click: function(event, target) {
           var position = self._getMenuPosition(target);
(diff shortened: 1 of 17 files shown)


lib/features/align-elements/AlignElementsContextPadProvider.js
```diff
@@ -70,7 +70,7 @@ AlignElementsContextPadProvider.prototype._getEntries = function() {
     'align-elements': {
       group: 'align-elements',
       title: self._translate('Align elements'),
-      imageUrl: ICONS['align'],
+      html: `<div class="entry">${ICONS['align']}</div>`,
```
barmac, line 73: Shouldn't it be this?
```suggestion
      imageHtml: `<div class="entry">${ICONS['align']}</div>`,
```

smbea: Actually, we used `imageHtml` to allow html for popup menu entries since the icon is just one element making up the entry. With context pad, this was already in place with `html`: https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/context-pad/ContextPad.js#LL320C32-L320C36 - we did not change this. Also for context pad, it represents the whole component since there is only the icon.

We could decide to change it to make it more consistent but would introduce a breaking change there.

barmac: OK now I get it. Thanks for the explanation.

marstamm (approved):
works for me 👍 

We are inconsistent with `imageHtml` and `html` as @barmac mentioned, but I think we can keep it for now and can consider aligning it at a later point

PR #1922, opened 2023-06-05
ci: extract translations update to separate workflow

barmac:
This allows to trigger translations update at convenience.

Related to https://github.com/bpmn-io/bpmn-js/actions/runs/5177347753/jobs/9327333594


nikku:
I like that we can also trigger this manually. Good stuff.

PR #1924, opened 2023-06-07
fix(ReplaceMenuProvider): Fix event subprocess replace menu not showing the correct entries

### LINKED ISSUE #1923: Replace menu doesn't show the corresponding non-interrupting element
### Describe the Bug

The replace menu does not show the option to replace by the corresponding non-interrupting element.

### Steps to Reproduce

1. select a "message start event"
2. open replace menu
3. see that "non-interrupting message start event" is not present
4. interrupting "message start event" still appears as an option

~~This happens in Camunda Modeler but not in the demo (see bello). For this reason, I believe this is a bug with the [UnlinkTemplateReplaceProvider](https://github.com/bpmn-io/bpmn-js-create-append-anything/blob/main/lib/element-templates/unlink-templates/UnlinkTemplateReplaceProvider.js) in this extension.~~ This element is also missing in `bpmn-js`.

<div>
Screenshot 2023-05-19 at 11 11 52

Screenshot 2023-05-19 at 11 11 52
</div>

### Expected Behavior

* The "non-interrupting" should appear as option
* The duplicate of the selected element should not appear

### Environment

 - Library version: latest

hkupitz:
capture gqbBtP_optimized

Closes #1923

### DIFF
--- a/test/fixtures/bpmn/features/replace/01_replace.bpmn
+++ b/test/fixtures/bpmn/features/replace/01_replace.bpmn
@@ -60,6 +60,9 @@
       <bpmn:startEvent id="StartEvent_3" isInterrupting="false">
         <bpmn:messageEventDefinition />
       </bpmn:startEvent>
+      <bpmn:startEvent id="StartEvent_6">
+        <bpmn:messageEventDefinition id="MessageEventDefinition_14m0uvi" />
+      </bpmn:startEvent>
     </bpmn:subProcess>
     <bpmn:startEvent id="StartEvent_4">
       <bpmn:timerEventDefinition id="TimerEventDefinition_0la1q3n" />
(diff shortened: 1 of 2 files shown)


smbea (changes requested):
This is great, thank you for your contribution! I tested and it's working great. A couple things:

1. Could we shorten the commit message so it doesn't break? Example:

```diff
- fix(ReplaceMenuProvider): Fix event subprocess replace menu not showing the correct entries
+ fix(replace): show correct interrupting/non-interrupting events entries
```

2. I would love to add a 'should show corresponding "non-interrupting" event' test around [here](https://github.com/bpmn-io/bpmn-js/blob/develop/test/spec/features/popup-menu/ReplaceMenuProviderSpec.js#L941).

smbea (approved):
I added a few minor improvements based on my comment.

Thanks for you contribution!

PR #1939, opened 2023-08-10
Drop callback style APIs

nikku:
Proposal to drop the node-style APIs that we still ship with our tools.

We released modern APIs with 7.0.0 (three years ago) and send a warning since then.


barmac (approved):
Makes so much sense :D

PR #1940, opened 2023-08-10
chore: extract modeling feedback into dedicated module

nikku:
It is a UI component that shall be used in interactive environments only.


philippfromme (approved):
Makes sense. ✅

PR #1965, opened 2023-10-05
chore: fix participant typo

makepanic:
I just fat fingered searching for participant in the repo and noticed that there were occurrences with the same mistake :laughing:


nikku:
Good stuff, great find :medal_sports:

nikku (approved):

PR #1968, opened 2023-10-09
ci: implement automated release

barmac:
Related to https://github.com/bpmn-io/internal-docs/issues/639

### DIFF
--- a/.github/workflows/POST_RELEASE.yml
+++ b/.github/workflows/POST_RELEASE.yml
@@ -1,8 +1,8 @@
 name: POST_RELEASE
 on:
-  push:
-    tags:
-      - 'v[0-9]+.*'
+  - workflow_call
+  - workflow_dispatch
+
 jobs:
   post-release:
 
--- a/.github/workflows/RELEASE.yml
+++ b/.github/workflows/RELEASE.yml
@@ -1,0 +1,50 @@
+on:
+  push:
+    branches:
+      - master
+      - develop
+name: release
+jobs:
+  release-please:
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+    permissions:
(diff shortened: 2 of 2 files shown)


barmac:
@nikku One more thought: This will also create GitHub Releases. If it's not what we want to have, we can disable it. Personally, it doesn't bother me.

.github/workflows/POST_RELEASE.yml
```diff
@@ -64,8 +64,3 @@ jobs:
         BPMN_IO_EMAIL: ${{ secrets.BPMN_IO_EMAIL }}
         BPMN_IO_USERNAME: ${{ secrets.BPMN_IO_USERNAME }}
       run: tasks/stages/update-website
-  update-translations:
```
nikku, line 67: Is there a reason you moved this?

barmac: I just moved it back because there was really no reason to do this.

barmac:
> Why would we want a GitHub release? As long as it will be created consistently, without work from our end, I think we could have it.

They are created automatically by release-please. I did not do anything specifically to create those at https://github.com/bpmn-io/dmn-variable-resolver/releases

I can see some arguments to start using GH releases:

* If one follows the repository on GitHub, they will be notified about the releases. Unlike blog posts, we don't need to write it manually.
* Disabling the feature is additional code.

I'd say the only argument against this is that we have not used GH releases outside of Camunda Modeler, i.e. tradition.

nikku:
Notifications + collecting :heart: should be enough in 2023 to start :see_no_evil:, let's do it.

nikku (approved):

barmac:
This is clearly not working right now. I will look into how we can improve the situation.

PR #1980, opened 2023-10-18
ci: fix POST_RELEASE job

### LINKED ISSUE #1979: Fix post-release job
### What should we do?

Fix post-release job which fails since automatic release was introduced. It fails to read the tag which breaks the pipeline.

Cf. https://github.com/bpmn-io/bpmn-js/actions/runs/6557498817/job/17809291824

### Why should we do it?

Our CI should work.

barmac:
Closes #1979

We are using release-please action outputs in this solution: https://github.com/google-github-actions/release-please-action#outputs

### DIFF
--- a/.github/workflows/POST_RELEASE.yml
+++ b/.github/workflows/POST_RELEASE.yml
@@ -1,7 +1,17 @@
 name: POST_RELEASE
 on:
-  - workflow_call
-  - workflow_dispatch
+  workflow_call:
+    inputs:
+      tag:
+        type: string
+        description: 'The released tag'
+        required: true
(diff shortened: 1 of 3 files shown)


barmac:
Manual test is running at https://github.com/bpmn-io/bpmn-js/actions/runs/6557829390

barmac:
POST_RELEASE works. There was a problem with translations update, but I could reproduce it only when called from the POST_RELEASE action.

barmac:
@philippfromme @marstamm ready for review.

marstamm (approved):
Looks like it works, great to have it fixed 🚀

PR #1982, opened 2023-10-19
ci: correctly output tag

### LINKED ISSUE #1979: Fix post-release job

barmac:
Closes #1979

This should work...


marstamm (approved):
🤞

PR #2000, opened 2023-11-01
feat(modeling): keep interrupting config when replacing elements

marstamm:
This PR adds 2 new features:

 - Keep the interrupting state when replacing elements, when the interrupting state is not explicitly set. This is used when element-templates are applied to events  
Recording 2023-11-01 at 13 24 20

- Add a toggle header button to non-interrupting events
image

related to https://github.com/bpmn-io/bpmn-js-element-templates/issues/20
You can try out the integration with element templates here: https://marstamm.github.io/non-interrupting-modeling-demo/

### DIFF
--- a/lib/features/modeling/behavior/NonInterruptingBehavior.js
+++ b/lib/features/modeling/behavior/NonInterruptingBehavior.js
@@ -1,0 +1,41 @@
+import CommandInterceptor from 'diagram-js/lib/command/CommandInterceptor';
+import inherits from 'inherits-browser';
+
+import { canBeNonInterrupting, getInterruptingProperty } from './util/NonInterruptingUtil';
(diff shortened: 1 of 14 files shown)


marstamm:
We could decide if it is a breaking change that `bpmnReplace.replaceElement()` now requires `cancelActivity` and `isInterrupting` to be explicitly set, when previously it defaulted to `true`

barmac:
Do you know if it's breaking any extension? I'd say that the previous behavior was bug-like. If I want to change type of my element, I shouldn't implicitly change other attributes.

marstamm:
I don't think it is breaking any extensions. Let's consider the previous behavior a bug and make it a minor release

barmac (approved):

PR #2026, opened 2023-11-15
fix(outline): end event

### LINKED ISSUE #2015: Outline around end events too narrow
### Describe the Bug

With the latest bpmn-js release we shipped outlines that stick to the elements. The one for end-events seems to be too narrow and this disturbs me:

capture XD4Azd_optimized

### Steps to Reproduce

1. Model start, intermediate and end event
2. Compare outline offset
3. See that end events have a much smaller offset than the other elements

### Expected Behavior

All outlines have same offset to shape.

### Environment

 - Browser: Chrome
 - OS: Linux
 - Library version: v15.1.0

smbea:
Closes #2015

This happens because the events all have the same diameter but end even has a thicker stroke.

### DIFF
--- a/lib/features/outline/OutlineProvider.js
+++ b/lib/features/outline/OutlineProvider.js
@@ -86,6 +86,19 @@ OutlineProvider.prototype.getOutline = function(element) {
       height: element.height + DEFAULT_OFFSET * 2
     }, OUTLINE_STYLE));
 
+  } else if (is(element, 'bpmn:EndEvent')) {
+
+    outline = svgCreate('circle');
+
+    // Extra 1px offset needed due to increased stroke-width of end event
+    // which makes it bigger than other events.
+
+    svgAttr(outline, assign({
+      cx: element.width / 2,
+      cy: element.height / 2,
+      r: element.width / 2 + DEFAULT_OFFSET + 1
+    }, OUTLINE_STYLE));
+
   } else if (is(element, 'bpmn:Event')) {
     outline = svgCreate('circle');
 


lib/features/outline/OutlineProvider.js
```diff
@@ -86,6 +86,16 @@ OutlineProvider.prototype.getOutline = function(element) {
...
     }, OUTLINE_STYLE));
 
+  } else if (is(element, 'bpmn:EndEvent')) {
+
+    outline = svgCreate('circle');
+
+    svgAttr(outline, assign({
+      cx: element.width / 2,
+      cy: element.height / 2,
+      r: element.width / 2 + DEFAULT_OFFSET + 1
```
barmac, line 96: Let's add a comment why `+ 1` is necessary:

> Increased stroke-width of end event grows both inward and outward which makes end event bigger than other events.

barmac:
What we have found is that the end event is bigger than other events due to how browsers render stroke. The solution proposed in this PR is to keep that difference in size, and adjust the outline. I played a bit with adjusting the size of the end event instead, but then run into problems (the `visual` size does not match anymore):

https://github.com/bpmn-io/bpmn-js/assets/28307541/5f0933db-b9e4-4573-a7ca-bc44a6f2fe21

So in the end I think it's OK to merge this, but please add an explanatory comment like the one I suggested in https://github.com/bpmn-io/bpmn-js/pull/2026/files#r1395412055

lib/features/outline/OutlineProvider.js
smbea: done

barmac (approved):

PR #2029, opened 2023-11-21
chore(CHANGELOG): update to v15.1.4

smbea:


smbea:
Closed since it will be taken care in https://github.com/bpmn-io/bpmn-js/pull/2030

PR #2036, opened 2023-11-27
fix(BpmnRenderUtil): use correct types

### LINKED ISSUE #2028: Type errors in BpmnRenderUtil.d.ts
type Shape is not defined and imported in the file which is being used in some methods in file.

I am importing a method from this file when I am trying to implement a custom renderer in a web based modeler built on anular framework, whole application is breaking , because Shape is not defined, can you please look into this.

import { getRoundRectPath } from 'bpmn-js/lib/draw/BpmnRenderUtil'; 
this is import where it is failing

### Steps to Reproduce

1. do this
2. do that

### Environment

 - OS: linux-ubuntu
 - Library version: latest- 15.1.3

marstamm:
closes #2028

### DIFF
--- a/lib/draw/BpmnRenderUtil.js
+++ b/lib/draw/BpmnRenderUtil.js
@@ -220,7 +220,7 @@ export function getRectPath(shape) {
 /**
  * Get width and height from element or overrides.
  *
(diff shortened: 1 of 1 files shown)


philippfromme (approved):
🚀

PR #2039, opened 2023-12-01
fix: add `isForCompensation` on connect

### LINKED ISSUE #2038: Can't add Compensation Activity via append menu
### Describe the Bug

It is not possible to append a task to a Compensation Boundary Event (Compensation Activity).

### Steps to Reproduce

1. select Compensation Boundary Event
2. open append menu
3. try to add a task/compensation ativity

### Expected Behavior

We should allow to append any Task and modify the action to add `isForCompensation` property.

### Environment

 - OS: [e.g. Windows 7]
 - Library version: [e.g. 2.0.0]

smbea:
Closes #2038


nikku (dismissed):
I integration tested this and indeed it feels like an overall _less_ blocking UX, which is a major improvement.

What I found is a couple of quirks:

> When transforming to a compensation boundary event the connection turns into an undirected association, where I'd expect a directed association. Expected: We use directed association and the task turns into a compensation task.

capture qX9Taq_optimized

> Existing sequence flows are not removed, despite being illegal for compensation tasks. Expected: Remove all other existing connections.

> Expected: Preview reflects future association layout.

capture fGgdTQ_optimized

lib/features/rules/BpmnRules.js
```diff
@@ -1028,7 +1028,7 @@ function isOneTextAnnotation(source, target) {
 function canConnectAssociation(source, target) {
 
   // compensation boundary events are exception
-  if (isCompensationBoundary(source) && isForCompensation(target)) {
+  if (isCompensationBoundary(source) && is(target, 'bpmn:Activity') && !isEventSubProcess(target)) {
```
nikku, line 1031: This likely prevents connecting event sub processes to text annotations?

capture l7Mo9k_optimized

smbea:
> When transforming to a compensation boundary event the connection turns into an undirected association, where I'd expect a directed association. Expected: We use directed association and the task turns into a compensation task.

This was a bug fixed via https://github.com/bpmn-io/bpmn-js/pull/2039/commits/aa2fd4179b8d1d4571b192f40654436cac1d4e29

> Existing sequence flows are not removed, despite being illegal for compensation tasks. Expected: Remove all other existing connections.

✅

> Expected: Preview reflects future association layout.

`drawPreview` is being called so I'm still not sure what the issue here is. Will have to look into the connection preview to understand more of how it works. I accept pointers if you have them 😅

> This likely prevents connecting event sub processes to text annotations?

Nope, doesn't affect text annotations

lib/features/modeling/behavior/CompensateBoundaryEventBehaviour.js
```diff
@@ -43,18 +82,53 @@ export default function CompensateBoundaryEventBehaviour(eventBus, modeling) {
...
 
-  this.postExecute('connection.delete', function(context) {
-    var source = context.source,
-        target = context.target;
+  this.postExecute('shape.replace', function(context) {
+    var oldShape = context.oldShape,
+        newShape = context.newShape;
+
+    if (replacing) {
+      replacing = false;
```
nikku, line 109: What do we want to accomplish here, and why?

The way we'd usually transport such context is by passing it via `hints` to the modeling API.

lib/features/modeling/behavior/CompensateBoundaryEventBehaviour.js
```diff
@@ -10,31 +10,70 @@ import { hasEventDefinition } from '../../../util/DiUtil';
...
+      });
     }
   }
 
+  this.postExecute('connection.delete', function(context) {
+    var source = context.source,
+        target = context.target;
+
+    if (isCompensationBoundaryEvent(source)) {
+      console.log('delete');
```
nikku, line 57:
```suggestion
```

nikku (commented):
I really like the UX progress you made :tada: 

capture UoeTgS_optimized

Feels really good. Let's ignore the preview bug for now.

lib/features/modeling/behavior/CompensateBoundaryEventBehaviour.js
smbea: By removing the illegal connections on` connection.create`, it was causing issues when replacing. Likely because replacing involves removing the current connection and creating a new one. So with that I'm trying to pause the behaviour while replacing so it doesn't mess up the replace.

I am interested in finding a better way to do this. I tried to stick to API but ended up with this solution

nikku: We discussed today how you could react to `a task becomes a compensate activity` to remove incoming and outgoing flows.

This way you could simplify your implementation, and get rid of the dependency tracking.

smbea: Fyi: it isn't so simple after all. Because `a task becomes a compensate activity` after reacting to the `connection.create`, which is the problem.

Me and Maciej are looking into this together

smbea:
Agreed with @barmac that he would take a look at this

barmac:
If there are more than one outgoing sequence flows, we end up with incorrect diagram:

image

My proposal is to convert only if there is a single sequence flow, and otherwise remove all of the connections.

barmac:
Another edge case found ;): 

https://github.com/bpmn-io/bpmn-js/assets/28307541/e8072a40-b533-4f1f-8f39-5095c013de52

barmac:
There is more:

https://github.com/bpmn-io/bpmn-js/assets/28307541/4ad0aa60-ec7d-47ff-af61-212962cf7c3f

barmac:
However, I got the basic interaction to work:

https://github.com/bpmn-io/bpmn-js/assets/28307541/bdebd4d2-594e-4cff-96b7-174bf2c01d86

barmac:
There is more:

https://github.com/bpmn-io/bpmn-js/assets/28307541/6c57ad0a-e65d-4867-a61c-41212c796b0c

barmac:
The diagram above will validate with the XSD but is not a valid BPMN because there can be only a single compensation handler (notice that it's always singular in the spec):

> 10.7.1 Compensation Handler
> A compensation handler is a set of Activities that are not connected to other portions of the BPMN model. The
> compensation handler starts with a catch Compensation Event. That catch Compensation Event either is a
> boundary Event, or, in case of a Compensation Event Sub-Process, the handler’s Start Event.
> A compensation handler connected via a boundary Event can only perform “black-box” compensation of the original
> Activity. This compensation is modeled with a specialized Compensation Activity, which is connected to the boundary
> Event through an Association (see Figure 10.121). The Compensation Activity, which can be either a Task or a
> Sub-Process, has a marker to show that it is used for compensation only and is outside the normal flow of the
> Process.

barmac:
There is more:


https://github.com/bpmn-io/bpmn-js/assets/28307541/e8e5246e-fb0e-4507-b1de-b0f844dca3e0

This is not correct BPMN. I believe we should disallow such connection.

barmac:
I've implemented some magic:


https://github.com/bpmn-io/bpmn-js/assets/28307541/24bdf345-3780-4b31-8534-11035f05ce0d

barmac:
Last bug to fix:

https://github.com/bpmn-io/bpmn-js/assets/28307541/a3026798-217b-4a0c-868e-0f0c26745858

barmac:
I believe I have fixed all of the remaining issues. We should test this extensively though.

barmac:
Test command: `npx @bpmn-io/sr bpmn-io/bpmn-js#fix-compensation`

barmac:
The connection preview is broken, but we can fix this as a follow-up. Here I am mostly interested in the interaction correctness.

Rules:
1. There can be only a single compensation activity
2. Compensation boundary event cannot have outgoing sequence flows

lib/features/modeling/behavior/CompensateBoundaryEventBehaviour.js
```diff
@@ -12,124 +12,105 @@ import { hasEventDefinition } from '../../../util/DiUtil';
...
   function removeIsForCompensationProperty(target) {
-    if (is(target, 'bpmn:Activity') && isForCompensation(target)) {
-      modeling.updateProperties(target, { isForCompensation: false });
-
-    }
+    modeling.updateProperties(target, { isForCompensation: undefined });
   }
 
-  function removeIlegalConnections(element) {
+  function removeIlegalConnections(element, exception) {
```
philippfromme: I would have expected something like `ignoredConnection(s)`. `Exception` is ambiguous. 💣

lib/features/modeling/behavior/CompensateBoundaryEventBehavior.js
```diff
@@ -94,23 +94,15 @@ export default function CompensateBoundaryEventBehavior(eventBus, modeling, bpmn
     modeling.updateProperties(target, { isForCompensation: undefined });
   }
 
-  function removeIllegalConnections(element, exception) {
```
philippfromme: 👏🏻

lib/features/rules/BpmnRules.js
```diff
@@ -1178,3 +1179,7 @@ function canCopy(elements, element) {
 function getRootElement(element) {
   return getParent(element, 'bpmn:Process') || getParent(element, 'bpmn:Collaboration');
 }
+
+function isHost(element) {
+  return element.attachers.length !== 0;
```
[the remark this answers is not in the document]
barmac: My assumption is that boundary event removal is too destructive and thus against "no magic" principle. How does it feel?

barmac:
> I think I found a bug. 🐛
> 
> brave_prMhl541BL [ brave_prMhl541BL ](https://private-user-images.githubusercontent.com/7633572/291829150-0cb0f15c-29c3-44bc-96d5-e547d0c7f5ed.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTEiLCJleHAiOjE3MDMwNjE0NTIsIm5iZiI6MTcwMzA2MTE1MiwicGF0aCI6Ii83NjMzNTcyLzI5MTgyOTE1MC0wY2IwZjE1Yy0yOWMzLTQ0YmMtOTZkNS1lNTQ3ZDBjN2Y1ZWQuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQUlXTkpZQVg0Q1NWRUg1M0ElMkYyMDIzMTIyMCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyMzEyMjBUMDgzMjMyWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MDdiN2I1NDQwODNlMGI5MWVjYTc2OGZjOTRkYjRlMDBmNjlhN2RlNzlhN2JjY2YxZjAwYTAwZGM2ZjFmZWYxOSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.fs0zwusSlkuDigevvoKF-px8vp5rGpfhgYymopjZ538) [ ](https://private-user-images.githubusercontent.com/7633572/291829150-0cb0f15c-29c3-44bc-96d5-e547d0c7f5ed.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTEiLCJleHAiOjE3MDMwNjE0NTIsIm5iZiI6MTcwMzA2MTE1MiwicGF0aCI6Ii83NjMzNTcyLzI5MTgyOTE1MC0wY2IwZjE1Yy0yOWMzLTQ0YmMtOTZkNS1lNTQ3ZDBjN2Y1ZWQuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQUlXTkpZQVg0Q1NWRUg1M0ElMkYyMDIzMTIyMCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyMzEyMjBUMDgzMjMyWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MDdiN2I1NDQwODNlMGI5MWVjYTc2OGZjOTRkYjRlMDBmNjlhN2RlNzlhN2JjY2YxZjAwYTAwZGM2ZjFmZWYxOSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.fs0zwusSlkuDigevvoKF-px8vp5rGpfhgYymopjZ538)

This is covered via https://github.com/bpmn-io/bpmn-js/issues/2061, and the interaction still produces a valid BPMN. We simply don't allow to configure association direction. It would be a different story if we had a dedicated "compensation association", but BPMN reuses regular association for the compensation activity.

lib/features/rules/BpmnRules.js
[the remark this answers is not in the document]
nikku: @barmac To be consistent with our existing UX we should remove the boundary event.

What we could do alternatively (this has been proposed by consultants once in a while) is to _detach_ the boundary event if possible, turning into an intermediate catch event. 

But as I see the later as a larger effort, I think for now the interaction should be _non -blocking_ all the way with users having the ability to undo.

nikku:
I did a manual test and could not break it. Regarding boundary event exception, I suggest to be [consistent in our UX](https://github.com/bpmn-io/bpmn-js/pull/2039#discussion_r1432413179).

lib/features/rules/BpmnRules.js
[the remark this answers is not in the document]
nikku: The alternative is that users have a way to learn about the _why_ :slightly_smiling_face:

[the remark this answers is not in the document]
barmac: Let's try this then. We could use modeling feedback too.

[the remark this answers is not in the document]
barmac: First step: remove attachments.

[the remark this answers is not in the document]
nikku: One thought for the future: We could also use `modeling feedback` to explain what we removed, and offer users a dedicated `undo` action. 

image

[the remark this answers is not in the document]
barmac: Getting there:

https://github.com/bpmn-io/bpmn-js/assets/28307541/66b0e7d1-3c34-4aa2-8dd8-1538680669a7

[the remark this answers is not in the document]
nikku: Removing a few too many right now :see_no_evil:

barmac:
I broke it:

https://github.com/bpmn-io/bpmn-js/assets/28307541/6a6032f6-f384-42bc-a1fc-0384e1c55852

barmac:
If we were to accept any reconnection, this should also be possible:

https://github.com/bpmn-io/bpmn-js/assets/28307541/bbfca8ed-07e1-4a96-af4d-474de1acd4fb

But I'd rather not implement it right now -> I'd say it's an unusual interaction.

barmac:
I implemented the boundary event removal, and also a couple of additional missing interactions. I believe this is mergable now, but please try to break it.

barmac:
Last bug (R) fixed via https://github.com/bpmn-io/bpmn-js/pull/2039/commits/4c82d9c4883c384f257ed8f198e391289b799d54

lib/features/rules/BpmnRules.js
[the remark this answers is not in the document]
barmac: > One thought for the future: We could also use `modeling feedback` to explain what we removed, and offer users a dedicated `undo` action.

I like this idea, but that could be a separate issue for all the cases we remove stuff. I'd rather not add it on top of this PR as it's already been quite a ride :D

barmac:
I retargetted this to `develop` as it's more a feature than a fix.

lib/features/modeling/behavior/CompensateBoundaryEventBehavior.js
```diff
@@ -0,0 +1,232 @@
...
+          oldSource = context.oldSource,
+          oldTarget = context.oldTarget;
+
+    // source changes
+    if (oldSource !== newSource) {
+      const target = oldTarget;
+
+      // newSource no longer valid
+      if (isForCompensation(target) && !isCompensationBoundaryEvent(newSource)) {
+        removeIsForCompensationProperty(target);
```
nikku, line 68: What scenario did we not cover here?

lib/features/modeling/behavior/CompensateBoundaryEventBehavior.js
```diff
@@ -0,0 +1,232 @@
...
+      }
+
+      // newSource valid
+      if (canBeForCompensation(target) && isCompensationBoundaryEvent(newSource)) {
+        addIsForCompensationProperty(target);
```
nikku, line 73: What scenario did we not cover here?

lib/features/rules/BpmnRules.js
```diff
@@ -550,17 +550,18 @@ function canConnect(source, target, connection) {
...
-  if (isCompensationBoundary(source) && isForCompensation(target)) {
-    return {
-      type: 'bpmn:Association',
-      associationDirection: 'One'
-    };
-  }
-
   if (canConnectAssociation(source, target)) {
 
+    if (isCompensationBoundary(source) && is(target, 'bpmn:Activity')) {
```
nikku, line 555: I still don't understand why we cannot have association and compensate association as dedicated steps. 

As we treat both as distinct, why not clearly represent that in code?

barmac: Makes sense. I will change that.

lib/features/modeling/behavior/CompensateBoundaryEventBehavior.js
barmac: I am sorry, I don't understand the question. Can you make it more verbose?

nikku: This line is not covered by tests, so I wonder which scenario this line represents that we do not not cover through tests.

nikku (approved):
This works great!

nikku:
From my point of view you can go ahead and merge this. Consider cleaning up the commits before.

lib/features/modeling/behavior/CompensateBoundaryEventBehavior.js
barmac: I will add test cases for that.

barmac: This allowed to detect another _last bug_.

image

barmac: However, this can recreated only when we forcefully reconnect start. It's not allowed by the rules.

barmac: I will leave it out of scope, and remove the respective source code.

lib/features/modeling/behavior/CompensateBoundaryEventBehavior.js
barmac: OK so I debugged it and source changes are handled via ReplaceConnectionBehavior. We don't need explicit compensation-specific code.

barmac:
image
image

Let's 🎃 it!

PR #2071, opened 2024-01-05
Exclusive Gateway Rendering without marker

### LINKED ISSUE #2063: Exclusive Gateway Rendering without marker

BPMN allows displaying exclusive gateways with or without an X shaped marker. This can be controlled using the `isMarkerVisible` attribute. However, the marker is always displayed, ignoring the presence (or absence) of the `isMarkerVisible` attribute.

Some modelers prefer to model without a marker, which is possible in other tools. Modelers ask why this isn't supported, readers who are used to the other style ask what the (missing) X means.

<details>
  <summary>Sample model</summary>

[isMarkerVisible.txt](https://github.com/bpmn-io/bpmn-js/files/13729033/isMarkerVisible.txt)
sample

</details>

Display the X marker when the `isMarkerVisible` attribute is set to true.
Hide the X marker when the `isMarkerVisible` attribute is either missing or set to false.

<details>
  <summary>Specification excerpts</summary>

bpmn202p289
bpmn202p400

</details>

You could display the marker even when the `isMarkerVisible` attribute is missing.
Pro: Activiti compatibility? See https://forum.bpmn.io/t/when-trying-to-view-a-bpmn-diagram-generated-using-activiti-6-0-0-beta1-the-bpmn-js-viewer-does-not-show-the-exclusive-gateways-as-expected-x/2486
Con: Divergence from the specification.

### Additional context

Possibly #221

In addition to the forum post above, there is another one indicating that this was once supported: https://forum.bpmn.io/t/x-symbol-for-xor-gateway-in-viewer/2342

sombrek:
Closes #2063

This was trickier than expected, because create, replace and append relied on the existing behavior.


lib/features/palette/PaletteProvider.js
```diff
@@ -75,12 +73,6 @@ PaletteProvider.prototype.getPaletteEntries = function() {
 
     function createListener(event) {
       var shape = elementFactory.createShape(assign({ type: type }, options));
-
-      if (options) {
-        var di = getDi(shape);
-        di.isExpanded = options.isExpanded;
-      }
```
barmac, line 82: This change is not related to the PR, is it?

lib/features/modeling/ElementFactory.js
```diff
@@ -211,7 +211,7 @@ ElementFactory.prototype.createElement = function(elementType, attrs) {
   }
 
   if (is(businessObject, 'bpmn:ExclusiveGateway')) {
-    di.isMarkerVisible = true;
+    attrs = applyAttribute(di, attrs, 'isMarkerVisible');
```
barmac, line 214: Let's set the attribute to a non-true value only if it's set explicitly (either to `undefined` or `false`) which can be checked via `attrs.hasOwnProperty('isMarkerVisible')`. This way we do not introduce a breaking change.

barmac (changes requested):
Hi,

Thank you for your contribution. I tested this and it works great. There is only a couple of changes that are necessary in order to merge this.

I tried this out together [bpmn-js-create-append-anything](https://github.com/bpmn-io/bpmn-js-create-append-anything/), and the change of the default `isMarkerVisible` value appears to be a breaking change:

https://github.com/bpmn-io/bpmn-js/assets/28307541/91b5ed0b-2b28-4191-afd7-8f69ba5731c1

While I understand that BPMN assumes no marker per default, I believe we should maintain the existing experience unless the library user wants to opt-in for no marker gateway. So `bpmn-js` should assume value `true` unless it is specified otherwise when calling the ElementFactory. Check out my code comment for a suggestion how to implement this.

lib/features/palette/PaletteProvider.js
sombrek: Note that I explicitly set `isMarkerVisible` to `true` for the action that is created here (same file), which affects `options`.
Until now, `options` seems to be tailored to `isExpanded` and that could break when I add code for `isMarkerVisible`. I noticed that removing this code didn't have any effect at all, so I opted for removing it altogether as a simplification.

I would argue it fits the PR, but I'll add it back if you think it's unrelated or even breaking other stuff.

lib/features/modeling/ElementFactory.js
sombrek: I can only _set_ `undefined` by excluding the attribute from the XML. As a result, we would incorrectly show the marker if the attribute is missing, but still hide it if the attribute were `false`.

Just to be clear, is that the intention?

sombrek:
@barmac Thank you for your feedback.

Please see my comments in the conversations to let me know how I should proceed.

lib/features/context-pad/ContextPadProvider.js
```diff
@@ -420,7 +420,8 @@ ContextPadProvider.prototype.getContextPadEntries = function(element) {
         'append.gateway': appendAction(
           'bpmn:ExclusiveGateway',
           'bpmn-icon-gateway-none',
-          translate('Append gateway')
+          translate('Append gateway'),
+          { isMarkerVisible: true }
```
barmac, line 424: This change won't be necessary if we implement what I suggested in https://github.com/bpmn-io/bpmn-js/pull/2071/files/10ca198b1d479805812eaa473d31b15a607358aa#r1447241633

lib/features/palette/PaletteProvider.js
barmac: OK it looks like the code you deleted is a duplication of https://github.com/bpmn-io/bpmn-js/pull/2071/files/10ca198b1d479805812eaa473d31b15a607358aa#diff-c5327c9e63a3ac25dea818bb475153d157e315968f4607b60ee6f3dba51b94c9R206

Let's keep it removed then.

lib/features/modeling/ElementFactory.js
sombrek: @barmac Thanks for the explanation.

Maybe I confused myself over this, but I fail to see how I can pass `isMarkerVisible: undefined`.

The property doesn't become available when using either the following code or trying the same from within the ContextPadProvider.

`elementFactory.createShape({ type: 'bpmn:ExclusiveGateway' });`
`elementFactory.createShape({ type: 'bpmn:ExclusiveGateway', isMarkerVisible: undefined });`
`elementFactory.createShape({ type: 'bpmn:ExclusiveGateway' }, { isMarkerVisible: undefined });`

So I don't have a chance to use `hasOwnProperty` there. Thus, I cannot differ `isMarkerVisible: undefined` from missing property.

sombrek: Never mind. I learned something about JavaScript and developer tools. Thanks.

sombrek:
@barmac I think the desired behavior is now implemented. The plugin continues to work with this code.

barmac:
I simplified the code a bit, and rebased this on top of `develop`. I will merge this as a `feat` as soon as the CI checks pass. Thank you once more for your excellent contribution.

PR #2074, opened 2024-01-09
fix: support core replace in compensation behavior

### LINKED ISSUE #2073: CompensateBoundaryEventBehavior.js: fail 'shape.replace' event handler when the `targetElement` is undefined
### Describe the Bug

There is a bug while using bpmn-js-element-templates with bpmn-js@16.3.0 after removing an element template:

```
EventBus.js:437 unhandled error in event listener TypeError: Cannot read properties of undefined (reading 'type')
    at handleReplacement (CompensateBoundaryEventBehavior.js:101:1)
```

It is caused by calling https://github.com/bpmn-io/bpmn-js-element-templates/blob/main/src/element-templates/cmd/RemoveElementTemplateHandler.js#L46 without `hints.targetElement`

As we can see from line https://github.com/bpmn-io/bpmn-js/blob/develop/lib/features/modeling/behavior/CompensateBoundaryEventBehavior.js#L95 the `targetElement` could be undefined.

As solution, we could check it for undefined before getting properties like `targetElement.type` or `targetElement.eventDefinitionType` on 100,101,115 lines

### Steps to Reproduce

1. apply some element template
2. remove element template

### Expected Behavior

Element template is removed without errors

### Environment

 - Library version: 16.3.1
 - bpmn-js-element-templates: 1.10.0

barmac:
Closes #2073

### DIFF
--- a/lib/features/modeling/behavior/CompensateBoundaryEventBehavior.js
+++ b/lib/features/modeling/behavior/CompensateBoundaryEventBehavior.js
@@ -92,13 +92,15 @@ export default function CompensateBoundaryEventBehavior(eventBus, modeling, bpmn
    * and remove the potential candidates for connection replacement to have a single compensation activity.
    */
   function handleReplacement(context) {
-    const { targetElement } = context.hints || {},
-          { oldShape } = context;
+    const {
+      newData,
+      oldShape
+    } = context;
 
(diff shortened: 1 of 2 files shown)


nikku (approved):
Good stuff.

Issue #2049, opened 2023-12-07
Allow text annotations to overlap with the borders of subprocesses and pools


CatalinaMoisuc:
### Context

Transferred from: https://github.com/camunda/web-modeler/issues/7213.
cc @till-stadtler

### Describe the problem that you are trying to solve

When adding or moving text annotations in a subprocess or a pool, it can happen that the border of the subprocess or pool is shifted. For me, this is often an undesired effect. 

Text annotations can become hidden behind subprocesses.

This behavior of text annotations can lead to subprocesses and pools that are larger than they need to be in regards to the relevant BPMN elements themselves.

### Describe the solution you'd like to see implemented

I would like text annotations to not affect the size of subprocesses or pools when adding or moving them around. This means they should be allowed to overlap with the borders of subprocesses and pools. (This already happens when the text annotation is "outside" the subprocess or pool. In this case, the text annotation becomes partially hidden. (This behavior is not consistent, regarding a text annotation being outside of a pool or not.))

Text annotations should always be visible, thus independent of the "scope" where they live in the XML (?).

I hope this change will allow BPMN processes to become more compact as it will be easier to shift text annotations to a desired location. Customers might choose to position text annotations outside of subprocesses or pools, which makes the BPMN process itself more compact, see below. Once this is implemented, a tool can be added to temporarily hide all text annotations to focus on the BPMN process itself.

To summarize, text annotations should be _always on top_ and not affect resizing of the containers they are in.

### Describe alternatives you've considered

Use ctrl + z to undo the undesired change.

### Additional Context

Adding a text annotation:
![new-text-annotation](https://github.com/camunda/web-modeler/assets/117343024/1cce255c-0936-4d56-a856-ab4dda0f9f63)

Moved the text annotation a bit to the right:
![moved-text-annotation](https://github.com/camunda/web-modeler/assets/117343024/be27a6ab-6081-4700-a078-25379a1ff1be)

Partially hidden text annotation:
![partially-hidden-text-annotation](https://github.com/camunda/web-modeler/assets/117343024/22da7a16-ce1b-4413-942c-741aa4dfd0f9)

I assume this requested handling of text annotations will lead the customers to design more compact in regards to relevant BPMN elements. Especially, if we implement a tool to temporarily hide text annotations. This way, text annotations do not heavily affect the size and compactness of the BPMN model.
[...]

PR #2078, opened 2024-01-11
feat: adjust TextAnnotations behavior

marstamm:
Recording 2024-01-11 at 13 57 23

- render Annotations always on top
- Disable auto-resize for text annotations

closes https://github.com/bpmn-io/bpmn-js/issues/2049

### DIFF
--- a/lib/features/modeling/behavior/TextAnnotationBehavior.js
+++ b/lib/features/modeling/behavior/TextAnnotationBehavior.js
@@ -12,6 +12,19 @@ export default function TextAnnotationBehavior(eventBus) {
 
   CommandInterceptor.call(this, eventBus);
 
+  // On Append, TextAnnotations will be created on the Root.
+  // The default for connections will create the connection in the parent of
+  // the source element, so we overwrite the parent here.
+  this.preExecute('connection.create', function(context) {
+    const { target } = context;
+
+    if (!is(target, 'bpmn:TextAnnotation')) {
+      return;
+    }
+
+    context.parent = target.parent;
+  }, true);
+
   this.preExecute([ 'shape.create', 'shape.resize', 'elements.move' ], function(context) {
     const shapes = context.shapes || [ context.shape ];
 
--- a/lib/features/modeling/behavior/TextAnnotationBehavior.js
+++ b/lib/features/modeling/behavior/TextAnnotationBehavior.js
@@ -1,0 +1,30 @@
+import inherits from 'inherits-browser';
+
+import { is } from '../../../util/ModelUtil';
+
+import CommandInterceptor from 'diagram-js/lib/command/CommandInterceptor';
+
+/**
+ * @typedef {import('diagram-js/lib/core/EventBus').default} EventBus
+ */
+
+export default function TextAnnotationBehavior(eventBus) {
+
+  CommandInterceptor.call(this, eventBus);
+
+  this.preExecute([ 'shape.create', 'shape.resize', 'elements.move' ], function(context) {
+    const shapes = context.shapes || [ context.shape ];
+
+    if (shapes.length === 1 && is(shapes[0], 'bpmn:TextAnnotation')) {
+      context.hints = context.hints || {};
+
+      context.hints.autoResize = false;
+    }
+  }, true);
+}
+
+inherits(TextAnnotationBehavior, CommandInterceptor);
+
+TextAnnotationBehavior.$inject = [
+  'eventBus'
+];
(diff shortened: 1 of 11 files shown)


nikku:
How does it feel?

marstamm:
> How does it feel?

Like it was supposed to work this way from the start :)

nikku (commented):
This already feels great!

I found one instance where text annotations are not like labels:

capture sgMQVj_optimized

They are still _dropped_ into an element, and accounted for as a child during resize from what it looks like.

What we can consider:

* Should the 100% behave like labels?

What we could consider:

* do we want to make any adjustments to copy/paste behavior?

capture hBxnqa_optimized

marstamm:
> They are still dropped into an element, and accounted for as a child during resize from what it looks like.

We can adjust that to be always rendered on the root element. This would solve the resize problem.
Recording 2024-01-11 at 16 01 34

> Should the 100% behave like labels?
> do we want to make any adjustments to copy/paste behavior?

This might be a bigger topic we'd want to discover separately. Labels have a 1..1 relation to an element, while annotations are 0..n and are only connected to the element via an association. For consistency with Message- and Sequence Flows, I would keep the selection explicit in what we want to copy.

image

nikku:
Agreed on both points! Let's render them always on the root, and on top (like labels) and keep copy/paste out of scope.

lib/features/modeling/behavior/TextAnnotationBehavior.js
```diff
@@ -12,6 +12,19 @@ export default function TextAnnotationBehavior(eventBus) {
 
   CommandInterceptor.call(this, eventBus);
 
+  // On Append, TextAnnotations will be created on the Root.
+  // The default for connections will create the connection in the parent of
+  // the source element, so we overwrite the parent here.
+  this.preExecute('connection.create', function(context) {
+    const { target } = context;
+
+    if (!is(target, 'bpmn:TextAnnotation')) {
```
nikku, line 21: Would this work inside collapsed sub-processes? I'd rather change the logic here to explicitly target the highest up `process`, `collaboration` or `sub-process` (which can store the annotation).

lib/features/modeling/behavior/TextAnnotationBehavior.js
```diff
@@ -12,6 +12,19 @@ export default function TextAnnotationBehavior(eventBus) {
 
   CommandInterceptor.call(this, eventBus);
 
+  // On Append, TextAnnotations will be created on the Root.
+  // The default for connections will create the connection in the parent of
+  // the source element, so we overwrite the parent here.
+  this.preExecute('connection.create', function(context) {
```
nikku, line 18: An alternative here would be to fix this during modeling interaction (hover), cf. https://github.com/bpmn-io/bpmn-js/blob/aa158e2d5a7de86b8a388945bbb56f217922f1ae/lib/features/modeling/behavior/FixHoverBehavior.js#L24. 

This is where lane drop targets are currently adjusted to the containing participant: 

capture A8sibz_optimized

marstamm: This is not only relevant during hover/manual connecting, but also when you append a text annotation via the context pad

nikku: Then this seems to be the sensible solution then. I'd still consider the hover fix to indicate where the thing will be dropped (no auto-append).

lib/features/modeling/behavior/TextAnnotationBehavior.js
marstamm: Yes, this works inside a collapsed subproces. The Annotation is always added the "plane" of the original target (cf. root element, cf. https://github.com/bpmn-io/bpmn-js/blob/b5bae70a64fde3c114e06b9a4850d6c7f6190295/lib/features/ordering/BpmnOrderingProvider.js#L146-L152

By picking the highest flownode container, we get the resize behavior you experienced before. I think it is easier and cleaner to render on the root, rather than hack the resize behavior to ignore annotations

lib/features/modeling/behavior/TextAnnotationBehavior.js
marstamm: I will look into this

marstamm: Adjusted hover behavior in [7a43c0e](https://github.com/bpmn-io/bpmn-js/pull/2078/commits/7a43c0e52247549ccf3697553c7a9f3f17b5885c)
Recording 2024-01-12 at 10 16 42

nikku (approved):
Works for me! Good job.

PR #2102, opened 2024-02-07
Rendering of Gateway Without Marker

### LINKED ISSUE #2063: Exclusive Gateway Rendering without marker

sombrek:
Closes #2063 

I think it was a mistake to rely on `attrs`. Without it, this is the most simple solution I found. Please have a look.

### DIFF
--- a/lib/features/modeling/ElementFactory.js
+++ b/lib/features/modeling/ElementFactory.js
@@ -216,11 +216,9 @@ ElementFactory.prototype.createElement = function(elementType, attrs) {
   }
 
   if (is(businessObject, 'bpmn:ExclusiveGateway')) {
-    if (has(attrs, 'isMarkerVisible')) {
-      if (attrs.isMarkerVisible === undefined) {
+    if (has(di, 'isMarkerVisible')) {
+      if (di.isMarkerVisible === undefined) {
         di.isMarkerVisible = false;
-      } else {
-        attrs = applyAttribute(di, attrs, 'isMarkerVisible');
       }
     } else {
       di.isMarkerVisible = true;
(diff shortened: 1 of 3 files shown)


barmac:
Indeed, it makes sense now. So we can pass the attributes that are to be applied to DI via `attrs.di`.

barmac:
I changed the base of PR to `main` as it's a fix.

barmac (approved):

Issue #2058, opened 2023-12-14
Post release job does not operate on new tag


nikku:
### Describe the Bug

Post release actions should trigger on new tags. This allows us to cut releases manually (if need be), and execute follow-up actions consistently.

### Steps to Reproduce

1. Cut release via `np` or `npm publish`
2. See that `POST_RELEASE` workflow does not execute (has to be manually triggered)

### Expected Behavior

Post release activities execute on new tags, automatically.

### Environment

Not relevant.

PR #2123, opened 2024-03-22
Simplify POST_RELEASE and make work on tag again

nikku:
Closes https://github.com/bpmn-io/bpmn-js/issues/2058

Mirrors https://github.com/bpmn-io/dmn-js/pull/840


barmac:
My main concern about this is that right now we trigger post release job only after npm publication was done. In the proposed (and also legacy) flow, we rely on the publishing to finish in the meantime after tag was pushed to GitHub. So it's a race condition which used to fail a lot before the current flow was introduced.

barmac:
What I'd like to avoid is that the post release job fails after automated release.

How about we run the post release job on GitHub release trigger? This is the last step of the release flow so we should not run into old problems again.

barmac:
So this event: https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=published#release

nikku:
> My main concern about this is that right now we trigger post release job only after npm publication was done. 

Well, why would we want to do POST_RELEASE activities when the `npm release` failed? We do things like updating the website "this is the new thing to download" and I don't think it makes sense to decouple both.

barmac:
> Well, why would we want to do POST_RELEASE activities when the npm release failed? We do things like updating the website "this is the new thing to download" and I don't think it makes sense to decouple both.

I meant that the "push" trigger decouples stuff. The fact that commit was pushed does not mean that the release is available on npm. That's why we need to wait for the release to be available.

Anyway, let's try this out.

nikku:
I'm fine with waiting, if it is 

(1) as simple as https://github.com/bpmn-io/bpmn-js/blob/develop/.github/workflows/POST_RELEASE.yml#L30
(2) the job can be fine-tuned / restarted in case of an npm outage

PR #2125, opened 2024-03-30
feat: Update gateway's outline path

miyuesc:
When I reset the resise rule, if I adjust the size of the gateway node, the original implementation will cause an abnormal display of the outline.

Like this.

image

Now, will become like this

image

### DIFF
--- a/lib/features/outline/OutlineProvider.js
+++ b/lib/features/outline/OutlineProvider.js
@@ -59,15 +59,10 @@ OutlineProvider.prototype.getOutline = function(element) {
   }
 
   if (is(element, 'bpmn:Gateway')) {
-    outline = svgCreate('rect');
-
-    assign(outline.style, {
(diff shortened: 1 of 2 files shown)


philippfromme:
Thank you for your contribution. By default, gateways aren't resizable so using a `<rect>` and simply rotating it was the first choice. To merge your contribution we'd need to make sure that the margins remain the same and that the corners are rounded.

Compare the current outline

image

to the changed outline

image

miyuesc:
> Thank you for your contribution. By default, gateways aren't resizable so using a `<rect>` and simply rotating it was the first choice. To merge your contribution we'd need to make sure that the margins remain the same and that the corners are rounded.
> 
> Compare the current outline
> 
> image
> 
> to the changed outline
> 
> image

Sorry, I didn't take notice of the problem of the rounded corner.

PR #2151, opened 2024-04-27
fix: error info use tagged template

ChaominRuan:


nikku (approved):
Looks good. Thanks for migrating these over to tagged template literals!

Issue #2160, opened 2024-05-14
make the artifacts element support flowElements


evin2333:
I have customized an element custom:Group, which inherits bpmn:Group and contains a new property flowElements:FlowElements.
When I finished saving the data and reloading the xml, I found that the child element under this custom:Group was not rendering.

nikku:
@evin2333 I checked against the BPMN 2.0 meta-model and found that `bpmn:Group` does not support `flowElements`. Am I mistaken?

We'll not accept any contributions to add custom extensions in the core. Instead I recommend you to build your custom version of our BPMN editor to support your extensions there.

PR #2161, opened 2024-05-14
feat(import): make the artifacts element support flowElements

evin2333:
[
](https://github.com/bpmn-io/bpmn-js/issues/2160)


nikku:
Cf. https://github.com/bpmn-io/bpmn-js/issues/2160#issuecomment-2112698986.

PR #2171, opened 2024-05-27
fix(import): handle missing BPMNDiagram#plane

nikku:
Without this change any auxiliary `<bpmndi:BPMNDiagram />` without a plane would crash the importer.

Now the (broken) BPMN diagram is simply ignored during import.

----

Related to https://github.com/bpmn-io/bpmnlint/pull/135


barmac (approved):
Makes sense 👍

PR #2172, opened 2024-05-27
Gracefully handle missing `BPMNDiagram#plane` in legacy sub-process behavior

nikku:
Previously we'd scroll through all diagrams, assuming `plane` would always be set.

### DIFF
--- a/lib/features/drilldown/SubprocessCompatibility.js
+++ b/lib/features/drilldown/SubprocessCompatibility.js
@@ -57,20 +57,9 @@ SubprocessCompatibility.prototype._handleImport = function(definitions) {
     self._processToDiagramMap[diagram.plane.bpmnElement.id] = diagram;
   });
 
-  var newDiagrams = definitions.diagrams.reduce(
-    function(newDiagrams, diagram) {
-
-      const plane = diagram.plane;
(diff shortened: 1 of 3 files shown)


lib/features/drilldown/SubprocessCompatibility.js
```diff
@@ -57,11 +57,20 @@ SubprocessCompatibility.prototype._handleImport = function(definitions) {
...
+
+      if (!plane) {
+        return newDiagrams;
+      }
+
+      const createdDiagrams = self._createNewDiagrams(plane);
+      return newDiagrams.concat(createdDiagrams);
+    },
+    []
+  );
```
marstamm, line 73: We could replace the reduce with a filter + flatMap. In my opinion, this would make the code easier to understand:

```suggestion
  var newDiagrams = definitions.diagrams
    .filter(diagram => diagram.plane)
    .flatMap(diagram => self._createNewDiagrams(diagram.plane));
```

nikku: I'm fine to sprinkle some functional magic in.

marstamm (approved):
✈️

Issue #1676, opened 2022-06-07
Cannot undo element removal: `Cannot read properties of undefined (reading 'appendChild')`


nikku:
__Describe the Bug__

Removing [complex diagram](https://demo.bpmn.io/s/application-processing) and reverting that change via undo triggers the following error:

```
Cannot read properties of undefined (reading 'appendChild')
    at index.esm.js:28:17
    at it (index.esm.js:44:3)
    at mr._createContainer (GraphicsFactory.js:112:5)
    at mr.create (GraphicsFactory.js:136:15)
    at Vn._addElement (Canvas.js:866:29)
    at Vn.addShape (Canvas.js:888:15)
    at rf.revert (DeleteShapeHandler.js:88:10)
    at CommandStack.js:334:31
    at ih._atomicDo (CommandStack.js:377:5)
    at ih._internalUndo (CommandStack.js:330:8)
    at ih.undo (CommandStack.js:228:12)
    at Object.undo (EditorActions.js:80:20)
    at qp.trigger (EditorActions.js:170:24)
    at KeyboardBindings.js:77:21
    at EventBus.js:519:13
    at sr._invokeListener (EventBus.js:371:19)
    at sr._invokeListeners (EventBus.js:352:24)
    at sr.fire (EventBus.js:313:24)
    at Ia._keyHandler (Keyboard.js:104:35)
    at Ia._keydownHandler (Keyboard.js:86:8)
    at HTMLDocument.r (helpers.ts:87:17)
```

__Steps to Reproduce__

1. Open [this diagram](https://demo.bpmn.io/s/application-processing)
2. Select all
3. Remove contents
4. Undo
5. __See error in DEV tools__

__Expected Behavior__

Undo works reliably.

__Environment__

 - Browser: Chrome 100
 - OS: Linux (Arch)
 - Library version: bpmn-js@9.0.3, bpmn-js@9.2

MaxTru:
@nikku in currents planning we were not able to assign this. Are you able to tackle? If not, plaese move to backlog / repropose next time.

nikku:
Looking into this today and will move to backlog unless I#m able to crack this.

marstamm:
The Root cause is not the DataStore behavior but our `makeProcess` behavior and root replacing behavior. This also happens with comments placed outside of the Participant:

[...]

PR #2173, opened 2024-05-27
feat: render markers as part of djs-visual

marstamm:
This allows styling with `canvas.addMarker` and css, as described in our [examples](https://github.com/bpmn-io/bpmn-js-examples/blob/main/colors/README.md#option-3-color-via-marker--css-styling).
Requires https://github.com/bpmn-io/diagram-js/pull/906

related to https://github.com/camunda/camunda-modeler/issues/4307

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to [bpmn-js](https://github.com/bpmn-io/bpmn-js) are documen
 
 ___Note:__ Yet to be released changes appear here._
 
+* `FIX`: allow styling markers with `canvas.addMarker` and css ([#2173](https://github.com/bpmn-io/bpmn-js/pull/2173))
+* `CHORE`: render flow markers as part of `djs-visual` ([#2173](https://github.com/bpmn-io/bpmn-js/pull/2173))
+* `DEPS`: update to `diagram-js@14.7.0`
+
 ## 17.7.1
 
 * `FIX`: correct call activity outline ([#2167](https://github.com/bpmn-io/bpmn-js/issues/2167))
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -65,7 +65,7 @@ import Ids from 'ids';
 
 import { black } from './BpmnRenderUtil';
 
-var rendererIds = new Ids();
+var markerIds = new Ids();
 
 var ELEMENT_LABEL_DISTANCE = 10,
     INNER_OUTER_DIST = 3,
@@ -116,10 +116,6 @@ export default function BpmnRenderer(
       defaultStrokeColor = config && config.defaultStrokeColor,
       defaultLabelColor = config && config.defaultLabelColor;
 
-  var rendererId = rendererIds.next();
-
-  var markers = {};
-
   function shapeStyle(attrs) {
     return styles.computeStyle(attrs, {
       strokeLinecap: 'round',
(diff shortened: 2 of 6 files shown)


lib/draw/BpmnRenderer.js
```diff
@@ -1978,15 +1975,15 @@ export default function BpmnRenderer(
...
+            markerStart: marker(parentGfx, 'conditional-flow-marker', fill, stroke)
           });
         }
 
         // default marker
         if (sourceSemantic.get('default') && (is(sourceSemantic, 'bpmn:Gateway') || is(sourceSemantic, 'bpmn:Activity')) &&
             sourceSemantic.get('default') === semantic) {
           svgAttr(connection, {
-            markerStart: marker('conditional-default-flow-marker', fill, stroke)
+            markerStart: marker(parentGfx, 'conditional-default-flow-marker', fill, stroke)
```
nikku, line 1986: In a nutshell this renders markers locally, so as a result they can be styled with the element? We compromise by having a slightly bigger DOM tree, but I think that is fine.

nikku (commented):
I like this change; local markers (for coloring) seems to be a good thing to do, and it benefits the issue at hand (CSS styling), too. 

Still need to E2E test this against a larger diagram. Good job :clap:

lib/draw/BpmnRenderer.js
marstamm: Correct. We do not de-duplicate markers by type and color anymore. This allows each marker to be styles individually

marstamm:
I manually tested against `complex.bpm` (https://demo.bpmn.io/s/application-processing) and there was no noticeable performance decrease

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ All notable changes to [bpmn-js](https://github.com/bpmn-io/bpmn-js) are documen
 
 ___Note:__ Yet to be released changes appear here._
 
+* `DEPS`: update to `diagram-js@14.7.0`
```
nikku, line 9: Let's keep our existing order `FEAT` -> `FIX` -> `CHORE` -> `DEPS`

nikku: Let's also make this more descriptive: What problem do we solve? We now allow coloring of markers via CSS.

nikku: We could add https://github.com/bpmn-io/bpmn-js/pull/2173/files#r1617203732 as an additional `FIX` entry in the changelog.

marstamm: I added the descriptive message as a `FIX` and the technical description as a `CHORE` (how we render SVG is not that relevant to the user): [88e70a5](https://github.com/bpmn-io/bpmn-js/pull/2173/commits/88e70a5ab8c07ea45dab33f380fc78be0671dc29)

nikku: :+1:

nikku (approved):
I don't assume that rendering of these cryptic markers was ever done via userland CSS, hence I regard these changes as not breaking.

Good job :cactus:

nikku:
@marstamm Please verify the diagram-js update does not break dmn-js, and if it does bump diagram-js there, too.

Please also cut a release of this change (in bpmn-js), so users can have a clean (and compatible) install again.

marstamm:
I verified that the diagram-js update does not break dmn-js. I'll release bpmn-js soon, but old bpmn-js versions are also compatible with the new diagram-js version. 

Only reason some tests failed before the update was a CSS selector checking for a certain number of `paths` in a visual, which broke the assertion when we added the marker with an additional path. The behavior was still working as intended

marstamm:
I want to get https://github.com/bpmn-io/bpmn-js/issues/1676#issuecomment-2135203329 into the release

marstamm:
Released as v17.8.0

Issue #2210, opened 2024-07-31
Moving elements beyond the border of a pool is possible when moving it outside the viewport


StefanLerch:
### Describe the Bug
Restricting the movement of an element inside the "RulesProvider" module can be bypassed by moving an element outside the viewport. With this, it is for example possible to stack pools on top of each other, see GIF. 

### Steps to Reproduce

1. Create a new model with two pools
2. Moving the pools on top of each other does not work
3. Move the viewport so that one pool is barely visible 
4. Now move the second pool beyond the viewport, on top of the other pool
5. The pool can be dropped, even though it is normally not possible

Result: The pools can be stacked on top of each other.

Or:

1. Create a new model with a pool
2. Create a Startevent inside of it
3. Try to move the element outside the pool's border
4. Now move the element beyond the viewport

![chrome_SU1NUSx2UP](https://github.com/user-attachments/assets/884ff42d-f7d8-4e37-aad9-b1521056b223)

Result: The element is moved to the position. Even though the pool increases its size afterward to contain the element, I would still argue that this is not intended behavior because moving the element beyond the border of the pool is otherwise restricted.

### Expected Behavior

The current viewport should be considered when moving elements. Locally, I could fix this by preventing any movement beyond it. In the case provided above, I would expect, that I am unable to move the element beyond the viewport. The pool should only increase in size if I drop the element close to the edge of the pool, but not far beyond its borders.

### Environment

- Browser: Chrome v127.0.6533.73
- OS: Windows 11 Enterprise
- Library version: 17.6.4, als [e.g. 2.0.0] [demo.bpmn.io](https://demo.bpmn.io/new)

barmac:
Thanks for reporting this. 

I've tested a similar interaction in Excalidraw and Miro, and both of them allow to move the element past the viewport if one 
is fast enough. 

The problem of extending the participant by far via this trick looks like a bug to me as well. I wouldn't deem it critical though.

Adam-Thometz:
I could pick this one up

Adam-Thometz:
I already came up with a fix I just need permission to push my branch

nikku:
Please propose your changes [through a pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request).

Adam-Thometz:
I tried doing that but I have to push my branch and I get a 403 when I try to do that

barmac:
[...]

PR #2244, opened 2024-10-07
Fix issue with moving elements beyond border

Adam-Thometz:
[See original issue for context](https://github.com/bpmn-io/bpmn-js/issues/2210)

### Proposed Changes

Change one line of code in `lib/features/rules/BpmnRules.js` in the `canMove` function:
```
- if (!target)
+ if (target === undefined)
```

### Reasoning
If a user starts dragging an element, the target is undefined. If a user drags an element outside the canvas, the target is null. Since both cases are considered nullish, they will both return true (meaning the user can drop the element). We want the latter case to return false.

### Steps to reproduce
bpmn-demo

1. Wrap elements in a pool
2. Try dragging an element within the pool.
  a. If dragged within, should be successful
  b. If dragged outside the element but within the canvas, should fail
  c. If dragged outside the element AND outside the canvas, should fail

### DIFF
--- a/test/spec/features/snapping/BpmnCreateMoveSnappingSpec.js
+++ b/test/spec/features/snapping/BpmnCreateMoveSnappingSpec.js
@@ -715,6 +715,38 @@ describe('features/snapping - BpmnCreateMoveSnapping', function() {
 
   });
 
+  describe('dragging elements', function() {
+    var diagramXML = require('./BpmnCreateMoveSnapping.process.bpmn');
+
+    beforeEach(bootstrapModeler(diagramXML, {
+      modules: testModules
+    }));
+
+    it('should limit movement when attempting to drag outside the viewport', inject(function(elementRegistry, move, dragging, canvas) {
+
+      var task = elementRegistry.get('Task_1');
+      var viewbox = canvas.viewbox();
+      var initialX = task.x;
+      var initialY = task.y;
+
+      // when
+      move.start(canvasEvent({ x: task.x + task.width / 2, y: task.y + task.height / 2 }), task);
+
+      // Attempt to move far outside the viewport
+      var farAwayX = viewbox.x + viewbox.width * 2;
+      var farAwayY = viewbox.y + viewbox.height * 2;
+      dragging.move(canvasEvent({ x: farAwayX, y: farAwayY }));
+
+      dragging.end();
+
+      // then
+      expect(task.x).to.be.above(initialX, 'Task should have moved right');
+      expect(task.y).to.be.above(initialY, 'Task should have moved down');
+      expect(task.x).to.be.below(farAwayX, 'Task should not have moved as far right as attempted');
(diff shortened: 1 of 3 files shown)


barmac:
Hi,

Thank you for your PR. Can you please make sure the contribution meets the requirements in [CONTRIBUTING](https://github.com/bpmn-io/bpmn-js/blob/develop/.github/CONTRIBUTING.md#create-a-pull-request)?
Specifically, I'd like you to provide a test case to verify the issue is fixed.
The commit message should also adhere to [conventional commits guidelines](https://www.conventionalcommits.org/en/v1.0.0/), and end with `Closes #issueNumber`.

Adam-Thometz:
Thank you for the feedback @barmac I updated the commit message and added a test to `BpmnCreateMoveSnappingSpec.js`

barmac:
Thanks, I will have a look at this again today.

test/spec/features/snapping/BpmnCreateMoveSnappingSpec.js
```diff
@@ -715,6 +715,38 @@ describe('features/snapping - BpmnCreateMoveSnapping', function() {
...
   });
 
+  describe('dragging elements', function() {
+    var diagramXML = require('./BpmnCreateMoveSnapping.process.bpmn');
+
+    beforeEach(bootstrapModeler(diagramXML, {
+      modules: testModules
+    }));
+
+    it('should limit movement when attempting to drag outside the viewport', inject(function(elementRegistry, move, dragging, canvas) {
```
barmac, line 725: I've reverted the change to the rules, and the test still passes. What we'd like to achieve is that the test is failing without the change you introduced: https://github.com/bpmn-io/bpmn-js/pull/2244/files#diff-6031dc6e8e3d4e38aef3c1dda35e6ff836a6b6be1263157300284d0191a6d5d4R931
Otherwise, the test does not verify the fix as it is passing without any changes to the library code.

Adam-Thometz: Would it be possible to view some documentation on how the inject function works? Trying to figure out how to simulate the viewport bc in the test it does move outside but it isn't perceived as a viewport so it doesn't snap back if that makes sense.

nikku: @Adam-Thometz inject will execute the test within a previously bootstrapped bpmn-js instance, and allows you to access internal services.

Similar to this:

```javascript
const bpmnJS = new BpmnModeler();

const testFn = function(eventBus) { ... };

bpmnJS.invoke(testFn);
```

jarekdanielak:
@Adam-Thometz are you still working on it?

barmac:
I am closing this for now as there's no updates. @Adam-Thometz please feel free to comment if you want to pick this up again.

PR #2247, opened 2024-10-10
Ensure attributes are passed to createElement

lblacio-craft:
Updated the logic to pass the attributes directly to the create function as it expects both the type and the attributes. This simplifies the code and ensures better flexibility and maintainability.

**The issue**  :
When calling the createElement method, the source and target attributes 
were lost because they were not passed correctly. This fix ensures that 
these attributes are properly transmitted during the element creation, 
ensuring correct and complete behavior.

### DIFF
--- a/lib/features/modeling/ElementFactory.js
+++ b/lib/features/modeling/ElementFactory.js
@@ -163,7 +163,7 @@ ElementFactory.prototype.createElement = function(elementType, attrs) {
       throw new Error('no shape type specified');
     }
 
-    businessObject = this._bpmnFactory.create(attrs.type);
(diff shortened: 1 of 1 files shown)


barmac:
Please make sure you adhere to [the contributing guidelines](https://github.com/bpmn-io/bpmn-js/blob/develop/.github/CONTRIBUTING.md#create-a-pull-request), specifically provide a unit test which verifies the solution, and adjust the commit message to be a conventional commit. Also, we need you to sign the CLA to be able to merge the contribution.

nikku:
This will pollute the BPMN model.

Issue #2269, opened 2025-01-07
Copy + Paste and then Undo Collapsed Sub process throws JS Errors 


anandcoupa:
### Describe the Bug

Copy Paste and then Undo of Collapsed Sub process throws following JS Errors 
```unhandled error in event listener TypeError: Cannot read properties of null (reading 'children')```
related to `MoveConnectionHandler.js`

### Steps to Reproduce

1. Create a sample BPMN consisting of a collapsed subprocess
2. Do a copy paste of the collapsed subprocess
3. Notice that the paste operation is successful and if you navigate to newly pasted node, you see the underlying elements from the copied subprocess.
4. Perform undo (cmd+z or ctrl + z)

Actual:
- Newly pasted collapsed subprocess stays as is, JS error seen in the console

Please find the attached recording

https://github.com/user-attachments/assets/787104dd-3001-4bc5-b9f1-09f3d94e9388

### Expected Behavior

- No JS errors thrown 
- Undo functionality works as expected.

### Environment

 - Browser: Chrome 129
 - OS: MacOS
 - Library version: v17.11.1

abdul99ahad:
I can reproduce the error. Moving it to `ready`

barmac:
I can reproduce too. This crashes the editor.

abdul99ahad:
@jarekdanielak to have a look on this.

Delonge12345:
@jarekdanielak Hello! When are you going to fix this bug? It breaks moving elements, resize and whole UX(

jarekdanielak:
I unassigned myself from the issue, as won't be actively working on it. If anybody wants to pick it up, there is some investigation done in the linked PR: https://github.com/bpmn-io/bpmn-js/pull/2275

@Delonge12345 as this is now tagged `spring cleaning`, there is a chance me or someone else will pick this up mid-February.

Delonge12345:
@jarekdanielak Thank you! I hope you can fix this, as it's critical for us. Only refreshing (F5) helps. I tried to catch this error, but by the time it appears, it seems the schema is already broken(

Delonge12345:
@jarekdanielak Good afternoon! Were you able to fix the bug?

nikku:
Came across this issue via https://github.com/bpmn-io/bpmn-js/pull/2388, cross posting my assessment from there:

> In certain scenarios undo would fail with the following error:
> 
> ```
[...]

PR #2275, opened 2025-01-17
Collapsed subprocess paste & undo bugfix

jarekdanielak:
### Problem

When a collapsed subprocess with children is pasted, the children are first created on the canvas root and then moved into the new subprocess.

When this action is undone, editor crashes because it attempts to move an element to the parent which is already removed from the canvas.

This seems to be a problem with order of operations - our subprocess implementation should first create all elements, then move them and undo those in the opposite order, but it seems to revert move action for some elements which parent was already removed (`shape.create.revert`).

### Proposed solution

This pull request adds a test to cover this case (https://github.com/bpmn-io/bpmn-js/commit/569ce9facfa838d1da408b4465a391245a11687d) and proposes to move children into the new subprocess after all elements are created. 

Unfortunately, this causes a problem with text annotations and labels which are not getting copied now.

### Additional context

For more detailed debugging discussion see https://github.com/bpmn-io/diagram-js/pull/957#issuecomment-2598503452 and below.

Closes https://github.com/bpmn-io/bpmn-js/issues/2269

### Checklist


jarekdanielak:
I've spent some more time debugging this and updated the top comment with my latest conclusions.

The proposed solution of moving children elements after `elements.create` looks like the most sane way of implementing this behavior, but I still have some questions:

- In the current implementation of `SubProcessPlaneBehavior`, why is `shape.create` executed for each pasted subprocess child, even though they seem to be already available in the post-`shape.create` of the subprocess element?
- Why are all the children available in the context of `shape.create`, but not in the context of `elements.create`?

I'd like to come back to this in a paired debugging session. cc. @barmac @nikku

nikku:
@jarekdanielak It seems like this PR stalled, and we did not get time to look into it. I'm moving it (and the linked issue) to `backlog`.

PR #2282, opened 2025-02-17
feat: allow to replace between variants of typed events

barmac:
Related to https://github.com/camunda/camunda-modeler/issues/4523

### Proposed Changes

This allows to replace between start/intermediate catch/intermediate throw/end event of the same type (where allowed in BPMN).

https://github.com/user-attachments/assets/d10506f8-d771-4c3e-bab7-def0a5d19e45

### Checklist

### DIFF
--- a/lib/features/popup-menu/ReplaceMenuProvider.js
+++ b/lib/features/popup-menu/ReplaceMenuProvider.js
@@ -100,6 +100,9 @@ ReplaceMenuProvider.prototype.getPopupMenuEntries = function(target) {
 
   var rules = this._rules;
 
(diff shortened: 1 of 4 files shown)


nikku (approved):
This works like a charm:

capture dKAf3c_optimized

Kinda unexpected that we don't handle the choice via `BpmnRules`, but seems to be the usual thing in the replace options.

barmac:
> Kinda unexpected that we don't handle the choice via `BpmnRules`, but seems to be the usual thing in the replace options.

I agree. Implementing this would be a bigger refactoring though.

Issue #2226, opened 2024-08-30
Space and lasso tool not moving groups and text annotations inside a Collaboration


fatimagarmar:
### Describe the Bug

When using the Space Tool or the Lasso Tool in bpmn-js version 17.9.2, the tools do not move groups or text annotations when clicking inside a bpmn:Collaboration. However, all other elements are moved correctly. This issue appears to be specific to elements within a Collaboration.

### Steps to Reproduce

1. Open a BPMN diagram in the editor.
2. Add a bpmn:Collaboration with various elements, including groups and text annotations.
3. Select the Space Tool or Lasso Tool.
4. Click and drag to create space or select elements inside the Collaboration.
5. Observe that groups and text annotations are not moved, while other elements are moved as expected.

### Expected Behavior

All elements, including groups and text annotations, should be moved when using the Space Tool or Lasso Tool.

### Environment

 - Browser: Chrome Version 126.0.6478.127
 - OS: Windows 11 Pro
 - Library version: bpmn-js 17.9.2

![bug space tool](https://github.com/user-attachments/assets/f8232dd0-a1a1-4726-9250-9e43aa05d6a0)

---

Related to https://github.com/camunda/camunda-modeler/issues/4797

jarekdanielak:
The text annotation will not move with other elements unless it's also selected with the lasso. Same way it won't move if you move its parent element. It might be a problem when using lasso inside of collaboration. You can try to select elements inside the collaboration with lasso tool and shift click the annotation to move it all together.

As for the group, I do indeed see some inconsistencies when trying to lasso it, but could you please share a screen recording of your exact incorrect behavior so that we are on the same page?

fagarma:
> The text annotation will not move with other elements unless it's also selected with the lasso. Same way it won't move if you move its parent element. It might be a problem when using lasso inside of collaboration. You can try to select elements inside the collaboration with lasso tool and shift click the annotation to move it all together.
> 
> As for the group, I do indeed see some inconsistencies when trying to lasso it, but could you please share a screen recording of your exact incorrect behavior so that we are on the same page?


https://github.com/user-attachments/assets/b25cbd94-9de8-4d7a-b5ea-d53687f6f456


If I'm not mistaken the space tool should move everything inside the collaboration, but it won't move text annotations or groups.

jarekdanielak:
Space tool is local by default. Try using it while holding Shift key.

nikku:
[...]

PR #2285, opened 2025-02-19
fix: move text annotations when a participant is resized by space tool

jarekdanielak:
Related to https://github.com/camunda/camunda-modeler/issues/4172

### Proposed Changes

When using the space tool locally inside of a participant, it now moves text annotations that are enclosed by it.

Unlike labels, text annotations that are outside of the participant won't move even if their target is in the participant.

https://github.com/user-attachments/assets/03daff09-61d7-4f47-a848-82236ffd7e31

### Checklist

### DIFF
--- a/lib/features/space-tool/BpmnSpaceTool.js
+++ b/lib/features/space-tool/BpmnSpaceTool.js
@@ -2,7 +2,9 @@ import inherits from 'inherits-browser';
 
 import SpaceTool from 'diagram-js/lib/features/space-tool/SpaceTool';
 
-import { getBusinessObject, is } from '../../util/ModelUtil';
+import { getEnclosedElements, getBBox } from 'diagram-js/lib/util/Elements';
+
+import { getBusinessObject, is, isAny } from '../../util/ModelUtil';
 
 import { isHorizontal } from '../../util/DiUtil';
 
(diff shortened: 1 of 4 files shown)


jarekdanielak:
Looking for elements to move involves searching through root children at some point. I'm concerted about performance on very large diagrams. Should I?

barmac:
Is it much different on [complex.bpmn](https://github.com/bpmn-io/bpmn-js/blob/develop/test/fixtures/bpmn/complex.bpmn)?

jarekdanielak:
Looks smooth.

https://github.com/user-attachments/assets/2ff0f65a-34d2-4cfc-9bd4-e5cf9fa8e32b

nikku (changes requested):
I think the algorithm can be improved as per https://github.com/bpmn-io/bpmn-js/pull/2285#issuecomment-2668949270, can you check? 

If possible I'd like to add artifacts to the set of movable elements _before_ calculating adjustments in diagram-js.

jarekdanielak:
> If possible I'd like to add artifacts to the set of movable elements before calculating adjustments in diagram-js.

You are absolutely right. Such a simple change I overlooked. See now.

> Closes https://github.com/bpmn-io/bpmn-js/issues/2226?

Partially because the author also mentions moving groups.

nikku:
@jarekdanielak I think this can further, drastically be simplified, and fixed comprehensively:

You mention that it is clear we want the same behavior for groups, why not fix it for groups, too? Both are _artifacts_, and can be handled in the _exact_ same way. Check https://github.com/bpmn-io/bpmn-js/pull/2285/commits/d3e839473084e0c3439b6efff4b7f3311d5d6a13 where I simplified the overall behavior, but also support all _artifacts_, while also supporting all local resize scenarios (inside sub-process, for example).

Looking at your test, you add them to an existing test bed. Where does this lead to? Tests are naturally growing over time, and we will end up with massive suites, if we are not smart about separating things. What I did with https://github.com/bpmn-io/bpmn-js/pull/2285/commits/3b3f6f56a0ad3997576d5d1f4cfd5768ede92236 is to add a test be _specifically_ for artifacts, in different scenarios:

image

The tests oblige to our rule `#1`: Meaningful names, everywhere + `#2`: focused tests (+ playgrounds).

Checkout https://github.com/bpmn-io/bpmn-js/pull/2285/commits/6b34cab40882834d0ec1e8f6eacc7616d2953502 to see how I added a few basic utilities to make testing smooth (unclutter, so we can comprehend what is happening).

nikku:
Please check https://github.com/bpmn-io/bpmn-js/pull/2285#issuecomment-2672555642, accept or reject my changes, and squash the commits :slightly_smiling_face:

PR #2293, opened 2025-03-10
feat: render event subprocess icons

### LINKED ISSUE #50: Render start event marker in collapsed event subprocesses

sombrek:
Closes #50

Draw event icons for collapsed event subprocesses.
No icon is drawn when a start event is missing or when there's more than one start event (forbidden by the spec).

Reuses existing code with few minor adjustments.

Demo (spec on the right):

evsub-demo

[event-subprocess-icons.bpmn](https://github.com/sombrek/bpmn-js/blob/event-subprocess-icons/test/fixtures/bpmn/draw/event-subprocess-icons.bpmn) contains all possible icons.

### DIFF
--- a/lib/draw/BpmnRenderer.js
+++ b/lib/draw/BpmnRenderer.js
@@ -462,8 +462,8 @@ export default function BpmnRenderer(
       var pathData = pathMap.getScaledPath('EVENT_MESSAGE', {
         xScaleFactor: 0.9,
         yScaleFactor: 0.9,
-        containerWidth: element.width,
-        containerHeight: element.height,
+        containerWidth: attrs.width || element.width,
+        containerHeight: attrs.height || element.height,
         position: {
           mx: 0.235,
           my: 0.315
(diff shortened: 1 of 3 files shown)


barmac:
Hi @sombrek! Thanks for your contribution. 

> There is one bug: When the start event type changes, the icon is not updated. It's only updated when the collapsed subprocess is moved afterwards. I have no idea how to fix this. Do you?

The renderer is reactive in a sense that it redraws only the elements which it is prompted to redraw. The elements are passed by the Command Stack, and it receives them from the Command Handlers (e.g. https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/modeling/cmd/CreateShapeHandler.js#L65). I checked and the event subprocess is not passed when event is replaced, and that's why it's not redrawn.

image

So the solution would be to mark the event subprocess as dirty when its child start event's type changes.

barmac:
@sombrek do you still want to work on this PR? I am converting this to draft since some changes are still required, but feel free to comment/mark it as ready when you have an update.

sombrek:
@barmac I'd like to get the PR merged, but I do not see a way to fix the remaining issue.

How can re-drawing the parent diagram be triggered?
I've tried using `modeling.updateProperties(newShape, {})` in a `postExecuted` function when the event subprocess changes, but it had no effect.

barmac:
Hmm if a simple solution does not work, we could as well merge the pull request, and create an issue for a missing rendering update.

barmac:
Follow-up issue: https://github.com/bpmn-io/bpmn-js/issues/2297

barmac:
This is published in v18.4.0. Thank you @sombrek!

PR #2300, opened 2025-04-11
fix(layout): correct arrow direction for boundary event to task (#2299)

### LINKED ISSUE #2299: Incorrect direction of the arrow from the boundary event to the subsequent element.
### Describe the Bug

Incorrect direction of the arrow from the boundary event to the subsequent element.

### Steps to Reproduce
1. Open any business process (BP).  
2. Place a task element on the diagram and add a boundary event to the task.  
3. Add another task element at the same level as the boundary event.  
4. Click on the boundary event, then in the pop-up menu, select the arrow tool and drag it to the task element.
Image

You can check it here: https://demo.bpmn.io/s/start

### Expected Behavior

Image

### Environment

 - Browser: [ Chrome 134.0.6998.166]
 - OS: [MacOS Sequoia 15.3.2]
 - Library version: [18.4.0]

ayushdiagarwal:
### Proposed Changes

This PR fixes [issue #2299 ](https://github.com/bpmn-io/bpmn-js/issues/2299) where a sequence flow from a `bpmn:BoundaryEvent` to a `bpmn:Task` results in a **reversed arrow direction** due to a horizontal backstep in the generated waypoints.

#### What this does

- Adds a conditional layout shortcut to `BpmnLayouter#layoutConnection`
- Detects when a `BoundaryEvent` is connected to a `Task` (and properly attached)
- Returns a clean 2-point horizontal connection:
  - From the right edge of the BoundaryEvent
  - To the left edge of the Task

This prevents the extra backstep and ensures the arrowhead appears in the correct forward direction.

Previously, arrowheads would render like this (reversed and overlapping the task):

image

With this change, they now render as expected:

image

### Trade-offs

Running `npm run all` currently results in **16 test failures**. These seem to be due to:
- Cases where the layout expects different orientations (top, right, vertical)
- Some tests that check for multi-segment waypoint paths

I’ve kept this change **narrow and targeted**, but happy to improve it to preserve broader compatibility if you'd prefer a more general fix.

---

### Checklist

- [x] **Brief textual description** of the changes present
- [x] **Visual demo** attached (can add on request)
- [x] **Steps to try out**: Create a sequence flow from a `bpmn:BoundaryEvent` to a `bpmn:Task` and verify arrow points forward
- [x] Related issue linked via `Closes #2299`

---

Thanks so much for reviewing — excited to contribute and open to all suggestions

### DIFF
--- a/lib/features/modeling/BpmnLayouter.js
+++ b/lib/features/modeling/BpmnLayouter.js
@@ -146,6 +146,27 @@ BpmnLayouter.prototype.layoutConnection = function(connection, hints) {
     connectionEnd = getConnectionDocking(waypoints && waypoints[ waypoints.length - 1 ], target);
   }
 
+  // 🩹 Fix for issue #2299:
+  // When connecting from a BoundaryEvent to a Task, avoid backstep layout
+  // and return a clean 2-point horizontal connection
+  if (
+    is(connection, 'bpmn:SequenceFlow') &&
+    is(source, 'bpmn:BoundaryEvent') &&
+    is(target, 'bpmn:Task')
+  ) {
+    return [
+      {
+        x: source.x + source.width, // right edge of boundary
+        y: source.y + source.height / 2
+      },
+      {
+        x: target.x, // left edge of task
+        y: target.y + target.height / 2
+      }
+    ];
+  }
+
+
   if (is(connection, 'bpmn:Association') ||
       is(connection, 'bpmn:DataAssociation')) {
 


barmac:
Hi!
Thanks for your pull request. Could you please add a test for the fix so that we can verify it works? Cf. [contributing guide](https://github.com/bpmn-io/bpmn-js/blob/develop/.github/CONTRIBUTING.md#adhere-to-the-unit-test-style).

nikku (changes requested):
@ayushdiagarwal This fix is overly simplistic and does not account for our layouting strategy in place. As a result, it breaks [many different tests](https://github.com/bpmn-io/bpmn-js/pull/2300/checks):

image

Consider to root cause the issue first, starting from a reproducible test case, and then go a head and investigate a good fix.

To verify your solution works, execute locally:

```
npm run all
```

ayushdiagarwal:
@nikku Yes, I'm sorry for that, you're right. This is my first time creating a PR, so I’m still learning. I initially tried a slightly different approach, which only resulted in two test case failures. However, I realize that even two failing tests aren’t ideal. I wasn’t able to pinpoint the exact root cause, so I modified the code to avoid affecting other test cases.

barmac:
> I was having some trouble figuring out how to determine if a sequenceFlow (or an arrow connecting one component to another) is straight.

I'd expect it to have two waypoints, and they should have a single shared coordinate.

To proceed with the PR, I'd expect to have a failing test case first. This should drive the solution, not the other way round.

barmac:
I'm converting this to a draft, but feel free to mark it as ready for review when you come back to it.

nikku:
Closing due to lack of follow-up.

PR #2316, opened 2025-05-07
Selection outline is available by default from Viewer and not Modeler

### LINKED ISSUE #2302: Selection outline removed from viewer
### Describe the Bug

In [v18.0.0](https://github.com/bpmn-io/bpmn-js/blob/develop/CHANGELOG.md#1800), we removed the selection outline from the viewers without marking it as a breaking change. In the linked issue, we have once established that [the outline belongs to the viewer](https://github.com/bpmn-io/bpmn-js/issues/2135#issuecomment-2047363331). However, we proceeded with the removal via https://github.com/bpmn-io/bpmn-js/pull/2253, and even accidentally removed it from Modeler which was reverted via https://github.com/bpmn-io/bpmn-js/pull/2258.

Selection outline is part of the Viewer, and should be available per default. If a user wants to show a bare SVG, they can display the outcome of `Viewer.saveSVG` call.

### Steps to Reproduce

1. Open a viewer
2. Click an element

### Environment

 - OS: [e.g. Windows 7]
 - Library version: v18.0.0 and later

---

Related to https://github.com/camunda/web-modeler/issues/13893
Also https://camunda.slack.com/archives/C032H77434N/p1744791449811209

abdul99ahad:
Closes #2302 
## Viewer
viewer

#### Known bugs
- When I select the elements using `Cmd + Click`, some text is also selected along the way which is not the case with modeler.
- Flow elements are not selected when selected individually since they don't have outline.

viewer-flow-elements-not-outlined

## Modeler
modeler

- Modeler single element select. Can select flow elements but I guess it's not an outline.

modeler-single

### Steps to Try out

run `npm run start:viewer` to see outline is available in viewer
run `npm start` to see outline is available in the modeler (via viewer)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to [bpmn-js](https://github.com/bpmn-io/bpmn-js) are documen
 
 ___Note:__ Yet to be released changes appear here._
 
+* `FIX`: re-add outline in viewer and removed from modeler (outline can be accessed via viewer) ([#2302](https://github.com/bpmn-io/bpmn-js/issues/2302))
+
 ## 18.6.1
 
 * `FIX`: copy error, escalation, message and signal references when copying elements ([#2301](https://github.com/bpmn-io/diagram-js/pull/2301))
--- a/lib/Modeler.js
+++ b/lib/Modeler.js
@@ -34,7 +34,6 @@ import ReplacePreviewModule from './features/replace-preview';
 import ResizeModule from 'diagram-js/lib/features/resize';
 import SnappingModule from './features/snapping';
 import SearchModule from './features/search';
-import OutlineModule from './features/outline';
 
 var initialDiagram =
   '<?xml version="1.0" encoding="UTF-8"?>' +
@@ -189,7 +188,6 @@ Modeler.prototype._modelingModules = [
   ResizeModule,
   SnappingModule,
   SearchModule,
-  OutlineModule
 ];
 
 
--- a/lib/Viewer.js
+++ b/lib/Viewer.js
@@ -5,7 +5,7 @@ import DrilldownModdule from './features/drilldown';
 import OverlaysModule from 'diagram-js/lib/features/overlays';
 import SelectionModule from 'diagram-js/lib/features/selection';
 import TranslateModule from 'diagram-js/lib/i18n/translate';
-
+import OutlineModule from './features/outline';
 import BaseViewer from './BaseViewer';
 
 
@@ -69,6 +69,7 @@ Viewer.prototype._modules = [
   CoreModule,
   DrilldownModdule,
   OverlaysModule,
+  OutlineModule,
   SelectionModule,
   TranslateModule
 ];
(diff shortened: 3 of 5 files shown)


barmac:
> When I select the elements using Cmd + Click, some text is also selected along the way which is not the case with modeler.

Are you able to reproduce it with the library version from before the bug was introduced? If so, I wouldn't care about this now.

barmac:
Regarding commit message:

> feat: re-add outline in viewer

I'd consider this PR to be a fix which we can release via patch release.

barmac (approved):
This is exactly what is needed to fix the problem. Please make sure to adjust the commit message as I suggested before you merge the PR (probably best squash it with meaningful message).

nikku (commented):
Did we root cause _why_ this happened, cf. [ISSUE_PHASES](https://github.com/bpmn-io/internal-docs/blob/main/development/ISSUE_PHASES.md)? Adding the outline is what we _shall do only when we want the outline to be present, not to "fix this bug".

I suspect that _user-select: none_ in the CSS yields the same effect, without the outline being present.

abdul99ahad:
The issue seems to be with the `MultiSelectionOutline` in diagram-js.

May-07-2025 16-38-36

nikku:
@abdul99ahad I don't understand: Does `user-select: none` fix the issue or not?

barmac:
> Did we root cause why this happened

What do you mean by _this_? Why did we remove the outline in the first place?

abdul99ahad:
> When I select the elements using Cmd + Click, some text is also selected along the way which is not the case with modeler.

If you're talking about this issue, the `user-select: none` in the multiselect css solves the problem

nikku:
> > Did we root cause why this happened
>
> What do you mean by this? Why did we remove the outline in the first place?

Turns out I probably mis-understood the issue fixed here. What we do now is re-introduce the outline as a viewer module, which we accidentally introduced a long time in the past. However, it was never officially a viewer module. The viewer shall work without styles (outlined [here](https://github.com/bpmn-io/bpmn-js/issues/2135#issuecomment-2446077605) and [here](https://github.com/bpmn-io/bpmn-js/issues/2135#issuecomment-2444636726)).

See also [viewer starter example](https://github.com/bpmn-io/bpmn-js-examples/blob/main/starter/viewer.html).

You write in the linked issue closed:

> In [v18.0.0](https://github.com/bpmn-io/bpmn-js/blob/develop/CHANGELOG.md#1800), we removed the selection outline from the viewers without marking it as a breaking change. 

Because it was never officially part of the viewer.

> In the linked issue, we have once established that https://github.com/bpmn-io/bpmn-js/issues/2135#issuecomment-2047363331. 

The point is: Being "just a renderer" is valuable, too, i.e. to display overlays, simulations, etc, or simply to "display from BPMN" instead of "SVG". All these things you cannot do with `#saveSVG` output.

> Selection outline is part of the Viewer, and should be available per default. If a user wants to show a bare SVG, they can display the outcome of Viewer.saveSVG call.

This is not what it was intended to be. Viewer is barebones, yes, but offers interactivity, ability to add overlays etc. It does not assume outline to be there. If you _want_, then you can of course add it, in your custom viewer distribution.

I vote to revert this PR.

Issue #1709, opened 2022-08-12
Copy and paste between different browser windows


nikku:
__Is your feature request related to a problem? Please describe.__

As a user I want to copy and paste parts of my diagram between different browser/app windows.

Use case: Have multiple diagrams opened in different tabs, but also have diagrams opened in different bpmn-js based applications (Camunda Modeler, Camunda Web Modeler, bpmn.io demo).

__Describe the solution you'd like__

* [ ] If I copy parts of the diagram it is put into the system clipboard in a way that
* [ ] Pasting with the [canvas selected](https://github.com/bpmn-io/diagram-js/pull/659) pastes from the system clipboard, if it holds bpmn-js contents

__Describe alternatives you've considered__

None.

__Additional context__

The web platform provides clipboard API as well as `copy` events we may be able to hook into.

On the bpmn-js side we'd need to ensure our copy/paste implementation is side-effect free, and the copied tree is serializable.

nikku:
Investigating this issue we found a couple of copy and paste related issues: https://github.com/bpmn-io/bpmn-js/issues/1708.

nikku:
A poor mans version of this feature can be implemented following along the [bpmn-js-copy-paste-example](https://github.com/nikku/bpmn-js-copy-paste-example).

It has the following limitations:

* Not using system clipboard, but instead using on board diagram-js facilities
* Only copying to local storage / reading from local storage => in browser, cross-tab copy + paste solution

nikku:
We were not able to follow up on the required changes to make browser native focus handling + copy/paste available across all our toolkits (https://github.com/bpmn-io/internal-docs/issues/614#issuecomment-1220481826).

BricePissard:
It seems that the problem comes from the breaking change of passing to the bpmn-js version 11.0.0 using the 3rd party library to handle the keyboard: https://www.npmjs.com/package/keyboardevent-key-polyfill

This library is instantiated at the page load but not updated while changing tabs or windows.
The method `keyboardeventKeyPolyfill.polyfill()` should be called each time a BPMN Diagram is rendered (?).

The problem is not yet fixed, but it would be nice if someone from BPMN.io look at it more carefully.

nikku:
Implemented via [bpmn-js-native-copy-paste](https://github.com/nikku/bpmn-js-native-copy-paste).

PR #2324, opened 2025-05-28
feat: allow copy/paste between browser windows

comlit:
added ability to copy and paste bpmn elemtents between browser window or different applications using the systems clipboard

implementation based on [this](https://forum.bpmn.io/t/contributing-as-part-of-the-informaticup-2025/14510/13?u=fabian1)

closes https://github.com/bpmn-io/bpmn-js/issues/1709

### DIFF
--- a/lib/features/copy-paste/BpmnCrossWindowCopyPaste.js
+++ b/lib/features/copy-paste/BpmnCrossWindowCopyPaste.js
@@ -1,0 +1,103 @@
+
+/**
+ * @typedef {import('diagram-js/lib/core/EventBus').default} EventBus
+ * @typedef {import('../../model/Types').Moddle} Moddle
+ * @typedef {import('diagram-js/lib/features/clipboard/Clipboard').default} Clipboard
+ */
+
+import { isObject } from 'min-dash';
+
+/**
+ * Cross-window copy-paste
+ *
+ * @param {EventBus} eventBus
+ * @param {Moddle} moddle
+ * @param {Clipboard} clipboard
+ */
+export default function BpmnCrossWindowCopyPaste(
+    eventBus,
+    moddle,
+    clipboard
+) {
+    function testFunc() {
+        return true;
+    }
+
+    function createReviver(moddle) {
+
+        var elCache = {};
+
+        /**
+             * The actual reviewer that creates model instances
+             * for elements with a $type attribute.
+             *
+             * Elements with ids will be re-used, if already
+             * created.
+             *
+             * @param  {String} key
+             * @param  {Object} object
+             *
+             * @return {Object} actual element
+             */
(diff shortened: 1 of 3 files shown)


barmac:
The CI is failing right now. Can you have a look?

comlit:
yes we looked at the failing tests and identified it happening because of issues with asynchronous code in testing, which did not occur while testing locally.  We are on it👍

barmac:
Hi,
I've tried this out, and it worked fine for a basic case. However, once we dig deeper, there are the following issues:

* The module logs an error for any text in the clipboard which does not meet the expectations.
* There are even uncaught errors in case a proper JSON was present in the clipboard. After that, the feature stops to work.
* Naturally, you cannot copy and paste elements with custom moddle extensions to an instance without it. This is expected, but we could handle this gracefully (e.g. strip extensions).
* Obviously, the browser asks for a permission to use the clipboard. I think this should happen only when one tries to copy or paste, and not immediately when canvas is focussed.
* There is a problem with serialization when I try to paste an element with a linked root element (receive task with messageRef): `Failed to parse clipboard contents:  Cannot read properties of null (reading '$type')`

https://github.com/user-attachments/assets/3d88710e-96dc-49d1-bfe9-9753f9cf76fe

> I think this should happen only when one tries to copy or paste, and not immediately when canvas is focussed.

This requirement might be difficult to implement given the asynchronous nature of the API. This could be a reason to make the feature optional.

barmac:
Suggestion: Look up how [excalidraw handles cross-browser copy and paste](https://github.com/excalidraw/excalidraw/blob/7c58477382883026b65ca770f31b87ff153654f8/packages/excalidraw/data/image.ts#L49). You can use an existing [copy-to-image](https://github.com/barmac/bpmn-js-copy-as-image) service, and add descriptor embedding to it.

barmac:
Or easier: Adopt [the pattern of setting `type`](https://github.com/excalidraw/excalidraw/blob/c1415004005d74961d43f7bc80cd0a3f2e1ab0cb/packages/excalidraw/clipboard.ts#L167) on the text contents, and ignore anything that does not match the expectation

barmac (changes requested):
Cf. the comments above.

comlit:
Hey, thank you very much for your feedback. Unfortunatly the submission deadline for the informaticup (which was our motivation to tackle this issue in the first place) ends tomorrow. This means that we as a team unfortunatly have to focus on other university work. Nevertheless we would like to see this feature merged and will try to tackle the requested changes in our own time.

barmac:
All right! Thanks for notice.

nikku:
Implemented via https://github.com/nikku/bpmn-js-native-copy-paste.

PR #2328, opened 2025-06-16
WiP feat: link external label with its parent

jarekdanielak:
Related to https://github.com/camunda/camunda-modeler/issues/369

### Proposed Changes

Show a dotted link between an external label and its parent when it's selected.

link-labels

### To do

- [ ] Find the proper layer to draw the lines on for non-root planes (e.g. expanded subprocesses)
- [ ] Find a better "middle point" for sequence flows with complicated shapes (not a straight line)
- [ ] Add tests
- [ ] Make the line more delicate

### Try

Run `npm start` on this branch and try it out.

### Checklist

### DIFF
--- a/lib/features/label-link/LabelLink.js
+++ b/lib/features/label-link/LabelLink.js
@@ -1,0 +1,79 @@
+import { createLine } from 'diagram-js/lib/util/RenderUtil';
+import { queryAll as domQueryAll } from 'min-dom';
+import {
+  append as svgAppend,
+  remove as svgRemove
+} from 'tiny-svg';
+
+import { is, isAny } from '../modeling/util/ModelingUtil';
+
+/**
+ * @typedef {import('diagram-js/lib/core/EventBus').default} EventBus
+ * @typedef {import('diagram-js/lib/core/Canvas').default} Canvas
+ */
+
+/**
+ * @param {EventBus} eventBus
+ * @param {Canvas} canvas
+ */
+export default function LabelLink(eventBus, canvas) {
+
+  // TODO: The links are not shown in subprocesses. Find the proper layer.
+  const defaultLayer = canvas.getDefaultLayer();
+
+  eventBus.on('selection.changed', function({ newSelection }) {
+
+    removeAllLinks(defaultLayer);
+
+    newSelection
+      .filter(element => isAny(element, [ 'bpmn:Event', 'bpmn:SequenceFlow' ]))
+
+      // when both label and target are selected
+      .filter(element => !newSelection.some(i => i.labelTarget === element))
+      .forEach(element => {
+
+        const target = element.labelTarget || element.labels[0];
+
+        if (!target) return;
+
+        const points = [
+          getMiddle(element),
+          getMiddle(target)
+        ];
+
+        const line = createLine(
+          points,
+          {
+            class: 'bjs-label-link',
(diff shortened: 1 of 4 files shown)


lib/features/label-link/LabelLink.js
```diff
@@ -0,0 +1,79 @@
...
+        svgAppend(defaultLayer, line);
+
+      });
+  });
+}
+
+LabelLink.$inject = [ 'eventBus', 'canvas' ];
+
+// TODO: Works kinda weird with complicated sequence flows.
+function getMiddle(element) {
```
nikku, line 62: What you want to use is `LineIntersection#getApproxIntersection(waypoints, reference)` or a derivative. This gives you the closest anchor point, either on a bendpoint, or on the connection itself.

nikku (commented):
This looks great:

* self contained
* simple enough
* working well

Looking forward for the follow-up to get it into the tool. One thing to consider: Whether to make it an (optional) diagram-js feature. I see that other tools, i.e. dmn-js, may benefit from it, too.

jarekdanielak:
We're getting there...

✅ I made sure the line is drawn between the correct midpoints and stops at the element edge, using `diagram-js` utilities. 

The link line is oddly off in some cases, and I'm having trouble figuring out why. It works for some other, very similar cases. 

The problem comes from `getElementLineIntersection` for the label path. Perhaps `graphicsFactory.getShapePath` is not really meant to be used with labels?

weird-links-1

nikku (commented):
Gave this a user test. Findings:

* Does it make sense to show the `---` line only when the label is the single selection? It can otherwise get very noisy when multi-selecting:

  Image

* There could be some confusion on the dotted line, being taken as an association - Did we consider to make it a dashed line, and maybe using a different than black color, i.e. some dark gray? Cf. also [original design sketches](https://github.com/camunda/camunda-modeler/issues/369#issuecomment-498560267).

  Image

philippfromme:
Nit-pick: the commit says _with its parent_ but that's not the relationship. The element is not the parent. 😉

philippfromme:
I love this feature! The only thing I'd consider is to make it look less like it's part of the diagram and more like it's part of the UI, for example like this:

image

jarekdanielak:
Made some changes and improvements based on your feedback.

1. Changed the style of the line to a dashed, light blue (our secondary outline color).
2. Show only one link. Do not show them when multiple elements are selected.
3. The link now attaches either to the element or its outline, depending on what's selected.

outline-tweak

outline-tweak-demo

jarekdanielak:
CI is failing, cause it seems the path is calculated slightly differently depending on the platform. I attempted to fix it in https://github.com/bpmn-io/bpmn-js/pull/2328/commits/159ab24bd75c6b6f515d61621b7a121f84ac3280, but it seems I didn't. @philippfromme mind having a look at the failing Windows test?

jarekdanielak:
In https://github.com/bpmn-io/bpmn-js/pull/2328/commits/822129c3af3036793bc734f14901c2ca6d86f74b, I added a small utility to compare SVG paths, allowing for some tolerance, as it appears that Headless Chrome renders them slightly differently depending on the platform.

philippfromme (approved):
Thanks for going the extra mile. With all the improvements this is now a great feature! 🎖️

PR #2329, opened 2025-06-17
Rules for non-interrupting StartEvents

Buckwich:
Related to https://github.com/bpmn-io/bpmn-js-create-append-anything/issues/4

### Proposed Changes

Only allow non-interrupting start-events on event subprocesses. If you try to move the start event out of the event subprocess it gets replaced with a normal none start event.

### Try Out
```
npx @bpmn-io/sr bpmn-io/bpmn-js-create-append-anything#non-interrupting-start-events -l bpmn-io/bpmn-js#rules-for-non-interrupting-start-events
```
image

 

### Checklist

### DIFF
--- a/lib/features/rules/BpmnRules.js
+++ b/lib/features/rules/BpmnRules.js
@@ -584,6 +584,11 @@ function canDrop(element, target) {
     return true;
   }
 
+  // allow non-interrupting start-events only in event sub processes
+  if (is(element, 'bpmn:StartEvent') &&
+          !isInterrupting(element)) {
+    return isEventSubProcess(target);
+  }
 
   // disallow to create elements on collapsed pools
   if (is(target, 'bpmn:Participant') && !isExpanded(target)) {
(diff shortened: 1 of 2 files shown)


barmac:
This change goes in line with what we implement for the compensation events:

https://github.com/user-attachments/assets/7d440dbd-7c8d-4b72-b93d-ed5a09e6f571

However, as you can see on the capture, for other container-specific elements (message start event, cancel event) we take a different stance. Namely, we allow the user to perform the action, and correct the element type for them. It works great with the preview as you can notice for element move. In the create mode, the preview is sadly missing, but could be added.

Let's consider [_empower the user_](https://github.com/bpmn-io/design-principles?tab=readme-ov-file#empower-the-user) in this context. I think that this principle can be realised better if our tool allows to perform the action, but at the same time shows the consequences in the preview. WDYT?

Buckwich:
I think the behavior you describe is the default without these rules: `npx @bpmn-io/sr bpmn-io/bpmn-js-create-append-anything#non-interrupting-start-events`

PR #2365, opened 2025-11-20
fix(label-link): only draw links if elements are on the visible plane

jarekdanielak:
### Proposed Changes

Before drawing a link, make sure the element is on the current plane.

This prevents links from staying on the root plane after a subprocess is collapsed.

**Good behavior**

good-label-link

**Current, bad behavior**

bad-label-link

### Checklist

* [ ] Related issue linked via `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`

### DIFF
--- a/lib/features/label-link/LabelLink.js
+++ b/lib/features/label-link/LabelLink.js
@@ -77,7 +77,7 @@ export default function LabelLink(eventBus, canvas, graphicsFactory, outline, se
 
   eventBus.on('shape.changed', function({ element }) {
 
-    if (!isAny(element, ALLOWED_ELEMENTS)) {
+    if (!isAny(element, ALLOWED_ELEMENTS) || !isElementSelected(element)) {
       return;
     }
 
@@ -178,6 +178,10 @@ export default function LabelLink(eventBus, canvas, graphicsFactory, outline, se
 
   function getElementPath(element) {
     return graphicsFactory.getShapePath(element);
+  }
+
+  function isElementSelected(element) {
+    return selection.get().includes(element);
   }
 }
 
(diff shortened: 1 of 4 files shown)


jarekdanielak:
@philippfromme can you have a look on Windows if the two tests render a correct path?

- `should show label for event in expanded subprocess`
- `should show label for event in collapsed subprocess plane`

They are passing on Mac and Linux. I already introduced some tolerance for Windows, but now they seem to be off by quite a lot.

nikku:
I wonder if there is anything else we're missing:

capture 4nOwzI_optimized


#### Steps to reproduce

* I collapse the sub-process
* I undo (CMD + Z)
* I see a link, despite neither start event nor label selected

(On a positive side, this "recovers").

jarekdanielak:
Great catch @nikku! I'm able to reproduce it manually, but not in a test case (cf. https://github.com/bpmn-io/bpmn-js/pull/2365/commits/cd54481d6dc79e47a1c010ff008c5cfe5ef5896a). Any ideas why?

nikku:
It works independent of collapse/expand, it is related to undo/redo. Maybe that helps you?

jarekdanielak:
Right! Looking into this led me to a simpler solution for both problems - just make sure the element is selected.

I was not able to break it now.

philippfromme:
The two tests you mentioned are failing on Windows but they look okay in the browser:

image

image

Could it be that it's about the difference between selected vs not selected?

brave_qcdHLI7k96

jarekdanielak:
The path is supposed to look different depending on whether the label or the target is selected. ☹️

jarekdanielak:
I can see now!

The same label appears as 2 lines on Windows, while it's one line on macOS:

image

jarekdanielak:
It boils down to differences in rendering fonts between macOS and Windows. I adjusted the label to always be single line.

PR #2376, opened 2026-01-11
Fix label snapping inside participants and lanes

### LINKED ISSUE #2371: Alignment tools for labels don't work inside of participant/lane
### Describe the Bug

Image

### Steps to Reproduce

### Expected Behavior

### Environment

 - Browser: Chrome 148
 - OS: Any (Linux, Ubuntu 24)

rohanSSBJ:
### Proposed Changes 
Closes https://github.com/bpmn-io/bpmn-js/issues/2371 

### Problem 
Labels inside participants with lanes weren't snapping to sibling shapes. This happened because `FixHoverBehavior` changes the label's hover target to the root element, causing snap targets to be computed from root children instead of the participant's children. 

**Solution**: Modified `BpmnCreateMoveSnapping.addSnapTargetPoints` to detect when moving a label whose `labelTarget` is inside a participant. When detected, shapes from that participant are now included as additional snap targets

### Changes 
- Added label-aware snap target logic in `addSnapTargetPoints` 
- Added `getParticipantAncestor` helper function 

### Proof (local working)
 Recording 2026-01-11 160207

### DIFF
--- a/lib/features/snapping/BpmnCreateMoveSnapping.js
+++ b/lib/features/snapping/BpmnCreateMoveSnapping.js
@@ -182,11 +182,20 @@ BpmnCreateMoveSnapping.prototype.addSnapTargetPoints = function(snapPoints, shap
 
     if (participant && target !== participant) {
 
-      // Add snap targets from the participant
+      // Get snap targets from the participant
       var participantSnapTargets = this.getSnapTargets(shape, participant);
 
+      // Add mid points for shapes and their labels
       forEach(participantSnapTargets, function(snapTarget) {
+
+        // Add the shape's mid point
         snapPoints.add('mid', getMid(snapTarget));
+
+        // Also add the shape's label's mid point if it has one
+        // (labels are not children of participant, so we need to add them explicitly)
+        if (snapTarget.label) {
+          snapPoints.add('mid', getMid(snapTarget.label));
+        }
       });
     }
   }
--- a/lib/features/snapping/BpmnCreateMoveSnapping.js
+++ b/lib/features/snapping/BpmnCreateMoveSnapping.js
@@ -13,8 +13,6 @@ import { isExpanded } from '../../util/DiUtil';
 
 import { is } from '../../util/ModelUtil';
 
-
-
 import {
   asTRBL,
   getMid
--- a/lib/features/snapping/BpmnCreateMoveSnapping.js
+++ b/lib/features/snapping/BpmnCreateMoveSnapping.js
@@ -160,7 +160,7 @@ BpmnCreateMoveSnapping.prototype.addSnapTargetPoints = function(snapPoints, shap
         snapPoints.add('mid', getMid(connection.target));
       }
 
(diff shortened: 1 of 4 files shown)


nikku:
@rohanSSBJ Thank you for your contribution! Could you [add a test case](https://github.com/bpmn-io/bpmn-js/blob/d1ccb83299d334a84ec1289ca5d87a2125c28c29/test/spec/features/snapping/BpmnCreateMoveSnappingSpec.js) to verify your solution across different scenarios that you identified (and commented in code)?

rohanSSBJ:
@nikku Thanks for the feedback!

I’ve added test coverage for label snapping inside participants and refined the implementation accordingly.

While running the `npm test`, I also noticed an existing failing test in `LabelBoundsSpec` on develop and included a small fix for that as well.

The test cases passed locally

rohanSSBJ:
While addressing the remaining uncovered lines reported by Codecov, I revisited the lane traversal logic.

In bpmn-js, flow nodes are direct children of the participant, not lanes.
`lane.children` only contains nested sub-lanes, so this code path is unreachable in normal operation.

I removed this dead code instead of adding test cases, which also brings patch coverage above the required threshold.

rohanSSBJ:
Hi @barmac just a small heads-up that CI is currently waiting for approval

No rush at all — happy to wait or adjust anything if needed.

barmac:
Thanks, I should have a look at this soon.

lib/features/snapping/BpmnCreateMoveSnapping.js
```diff
@@ -13,6 +13,8 @@ import { isExpanded } from '../../util/DiUtil';
 
 import { is } from '../../util/ModelUtil';
 
+
+
```
barmac:
```suggestion
```

test/spec/features/snapping/BpmnCreateMoveSnappingSpec.js
```diff
@@ -485,6 +485,82 @@ describe('features/snapping - BpmnCreateMoveSnapping', function() {
...
+        // The fix in addSnapTargetPoints detects shape.labelTarget,
+        // finds the participant ancestor, and adds snap targets from it
+        move.start(canvasEvent(labelMid), startEventLabel);
+        dragging.move(canvasEvent({ x: labelMid.x + 50, y: labelMid.y }));
+        dragging.end();
+
+        // then - verify move completed without errors
+        // The code path through getParticipantAncestor and the
+        // label snapping fix was exercised
+        expect(startEventLabel.x).to.exist;
```
barmac: This and the following test assertions don't verify that the snapping has actually taken place. I tried to implement a proper test case, but it does not seem to work. @philippfromme perhaps you could help with that later this week? I noticed in the debugger that the `event.delta` is overwritten properly, but then the [Move](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/move/Move.js#L157) feature uses `event.context.delta` instead.

PR #2377, opened 2026-01-19
fix: null check connection `sourceRef`

jarekdanielak:
### Proposed Changes

A conditional expression is now reused for Conditional Events, which don't have `sourceRef`, so I added a null check to make this behavior compatible with the new element type.

Required for https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1176

### Checklist

### DIFF
--- a/lib/features/modeling/behavior/ReplaceConnectionBehavior.js
+++ b/lib/features/modeling/behavior/ReplaceConnectionBehavior.js
@@ -178,7 +178,7 @@ export default function ReplaceConnectionBehavior(eventBus, modeling, bpmnRules,
     }
 
     // remove default from source on change to conditional
-    if (properties.conditionExpression && businessObject.sourceRef.default === businessObject) {
+    if (properties.conditionExpression && businessObject.sourceRef?.default === businessObject) {
       modeling.updateProperties(element.source, { default: undefined });
     }
   });


nikku (changes requested):
Please attach a test case, cf. [tests](https://github.com/bpmn-io/bpmn-js/blob/6c927bc42c27232c37125f5a7100ee4cfe80a2ec/test/spec/features/modeling/behavior/ReplaceConnectionBehaviorSpec.js#L4) to show what you're fixing, why.

Buckwich:
_Is the title/commit message right? Or did you mean conditional_

Edit: seems to be right

jarekdanielak:
I'm not sure adding a test to the replace connection behavior spec makes sense. 🤔

This behavior is not supposed to apply any changes to the conditional event. Like any other behavior, its test suite doesn't cover the elements it's _not_ supposed to act on.

What would make sense, is having a test to verify that a condition can be set on a conditional event - such test would detect that this behavior unintentionally breaks it - I added such tests in the `bpmn-js-properties-panel` implementation [here](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1176/changes#diff-464cf0a2a77c9479239662e579e4787a25e1f00afc1e27d17cc52de967b95433R60).

Questions:
1. Should we have a test somewhere in the `bpmn-js` to verify the condition of a conditional event works?
2. Should we adjust the replace connection behavior to only act on explicitly configured element types, i.e. sequence and message flows, instead of fixing it with a null check?

barmac:
Can you share a use case where the fixed code breaks without the change? What is the root cause?

jarekdanielak:
In https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1176, a conditional start event has a `conditionExpression` property, but no `businessObject.sourceRef`.

Without the change, an error is thrown when trying to evaluate second part of this condition:

```js
if (properties.conditionExpression && businessObject.sourceRef.default === businessObject) {
  modeling.updateProperties(element.source, { default: undefined });
}
```
Because `sourceRef` is null.

This worked so far, because only a sequence flow (which always has `sourceRef`) fulfilled the first part of the condition.

barmac:
I fail to find that property in the PR (apart from [sequence flow](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1176/files#diff-66038b540184d9eed29d5987e5e0b66432b00a95f7bb36ad0eebda333c658bcdR21)), and in the schemas: [bpmn-moddle](https://github.com/search?q=repo%3Abpmn-io%2Fbpmn-moddle%20conditionExpression&type=code), [zeebe-bpmn-moddle](https://github.com/search?q=repo%3Acamunda%2Fzeebe-bpmn-moddle%20conditionExpression&type=code). The conditional event definition uses `condition`, not `conditionExpression`.

I think the changed code could break in case of a custom moddle extension, but we are not dealing with it right now.

jarekdanielak:
Bummer. ☹️ 

I must've reused `conditionExpression` earlier in the implementation, fixed the crash in `bpmn-js`, changed the property to `condition` and missed the mismatch.

We might still decide to adjust the behavior to be more explicit, but this "fix" is not needed.

barmac:
Well at least you are not dependant on the bpmn-js change :)

nikku:
Great to see this closed. When something is broken, there is 98.73% chance we can create a simple reproducible test case. Where that is not the case (in our realm) the test setup needs improving, or stuff is actually not broken :upside_down_face:.

PR #2388, opened 2026-02-25
fix: sub-process annotations handling during collapse/expand, copy/paste, and remove actions

AlekseyManetov:
### Proposed Changes

Added correct handling for sub-process annotations during collapse/expand, copy/paste, and remove actions. 

The root cause of this behavior is that 'Text Annotations' are placed at the `bpmn:process` level rather than inside the `bpmn:subProcess`. While they appear visually inside, they are not included in `element.children`. Consequently, copy/paste and collapse operations lose them because these operations are intended to work only with children. 

Based on the possible solution options [discussion](https://github.com/camunda/camunda-modeler/issues/5163#issuecomment-3946571271), manual handling for annotations was implemented for sub-process actions. Additionally, nested sub-processes were addressed by ensuring annotations are processed recursively across all levels.

**Before:**

https://github.com/user-attachments/assets/d49e300e-2a15-407f-9ae6-b456dbe8447d

**After:**

**Copy/Paste:**

https://github.com/user-attachments/assets/de227abd-cb1d-4507-ad37-34153430c9dd

**Collapse/Expand:**

https://github.com/user-attachments/assets/9575429e-49bf-4757-8701-5c023ac24564

**Remove/undo:**

https://github.com/user-attachments/assets/8a7a54e0-7791-4b92-8890-27d606c3d4e7

Related to https://github.com/camunda/camunda-modeler/issues/5163
Related to https://github.com/camunda/camunda-modeler/issues/4987

Closes https://github.com/bpmn-io/bpmn-js/pull/2275
Closes https://github.com/bpmn-io/bpmn-js/issues/2269

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to [bpmn-js](https://github.com/bpmn-io/bpmn-js) are documen
 
 ___Note:__ Yet to be released changes appear here._
 
+* `FIX`: correct handling of annotations during sub-process collapse/expand, copy/paste, and remove actions ([#2388](https://github.com/bpmn-io/bpmn-js/pull/2388))
+* `FIX`: allow undo of pasted sub-process ([#2388](https://github.com/bpmn-io/bpmn-js/pull/2388), [#2269](https://github.com/bpmn-io/bpmn-js/issues/2269))
+
 ## 18.13.1
 
 * `FIX`: correct sequence flow layout for corner boundary events whose target is strictly axis-aligned ([#2270](https://github.com/bpmn-io/bpmn-js/issues/2270))


ev-codes (commented):
Thanks! 

Because I'm still onboarding, I won't approve this, but I left some comments (mostly for my own learning).

AlekseyManetov:
> Interesting edge case: What if annotation links to sub-process and "in sub-process" elements

While we don't restrict this, I'm wondering if this is a real use case that we should address as well. 

The only good solution I see is to split such annotation associations: move the element's annotation inside the sub-process while leaving the sub-process's annotation at the root. However, this would result in two separate annotations when the sub-process is expanded back. I don't think we can do this in a non-destructive way, and it would likely make the logic a bit complicated.

@nikku, do we need to address this issue, or should we just leave it as a corner case if it's not a real-world scenario?

nikku:
> @nikku, do we need to address this issue, or should we just leave it as a corner case if it's not a real-world scenario?

We should address it to the point where the behavior is not broken. You could choose to (1) leave the annotation pointing to the sub-process, disconnecting local other uses or (2) do the other way around.

AlekseyManetov:
> We should address it to the point where the behavior is not broken. You could choose to (1) leave the annotation pointing to the sub-process, disconnecting local other uses or (2) do the other way around.

Fixed with (1) option, only leave the annotation connected to the sub-process. Also fixed "Collapsing sub-process includes linked annotation" bug, now it stays outside of sub-process after collapsing.

Next - look at "Cannot read properties of null"(https://github.com/bpmn-io/bpmn-js/issues/2269) problem

nikku (commented):
@AlekseyManetov did you give it a manual integration test (`npm start`), verifying your solution works as expected?

AlekseyManetov:
> @AlekseyManetov did you give it a manual integration test (npm start), verifying your solution works as expected?

Sure. The previous bug identified during the last review was fixed(before it was a bug when we have multiple connections for subprocess elements):

https://github.com/user-attachments/assets/f3cb3e55-b337-4460-a09e-9cc82d35ee9b

Also fixed a bug related to https://github.com/bpmn-io/bpmn-js/issues/2269 issue:

https://github.com/user-attachments/assets/c4697498-5b8a-4a28-9c4a-41c639ef724d


So PR is ready for final review now.

nikku:
Added https://github.com/camunda/camunda-modeler/issues/4987, which is related to this PR.

AlekseyManetov:
@nikku, since I see you added the 'Fixed upstream' label to [#5163](https://github.com/camunda/camunda-modeler/issues/5163), can I take this as an approval to merge it now?

nikku:
I'm actively looking into reviewing it - by moving connected elements to `fixed upstream` I indicated that an upstream fix is available, that should be merged and incorporated.

barmac:
So far I was not able to break it. I could not reproduce the issues reported previously by Nico.

PR #2401, opened 2026-03-21
fix(snapping): allow label snapping inside participants and lanes

### LINKED ISSUE #2371: Alignment tools for labels don't work inside of participant/lane

sahilforkshere:
Fixes #2371

Root Cause: > FixHoverBehavior.js:49 forces the hover target to the root element whenever a label is being moved. In a collaboration diagram, the root is the collaboration element whose direct children are only participants — not the gateway/event shapes inside them. So getSnapTargets returned an empty set for labels inside participants, making alignment snap lines invisible.

Fix: > In BpmnCreateMoveSnapping.js, the getSnapTargets method now detects when the moving shape is a label whose actual parent differs from the forced root target (shape.labelTarget && shape.parent !== target). In that case, it additionally collects siblings from the label's real parent container (the lane/participant), merging them into the snap target list. This gives the snapping system access to the label's siblings (other labels, the labelTarget gateway/event, etc.), restoring the alignment helpers inside participants and lanes.

### DIFF
--- a/lib/features/snapping/BpmnCreateMoveSnapping.js
+++ b/lib/features/snapping/BpmnCreateMoveSnapping.js
@@ -3,6 +3,7 @@ import inherits from 'inherits-browser';
 import CreateMoveSnapping from 'diagram-js/lib/features/snapping/CreateMoveSnapping';
 
 import {
+  getChildren,
   isSnapped,
   setSnapped,
   topLeft,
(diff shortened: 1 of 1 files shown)


jarekdanielak:
I rebased your branch on the latest develop, cause we've recently introduced resizable external labels (https://github.com/bpmn-io/bpmn-js/pull/2414), and it seems to mess with the snapping, too.

Would you be interested in continuing the investigation to fix it once for all? 😄 

bpmn-js-snapping-1

PR #2402, opened 2026-03-21
Fix alignment tools working inside participant/lane

### LINKED ISSUE #2371: Alignment tools for labels don't work inside of participant/lane

qianchongyang:
## Problem

Alignment helpers (snap lines) work correctly for external labels outside of a participant, but fail to provide alignment assistance when the element is inside a participant or lane. The label target is not recognized as a "peer" for horizontal center alignment.

## Solution

Updated the alignment logic to correctly identify label targets inside participants and lanes. The snap line calculation now consistently handles elements regardless of their modeling context.

## Validation

- External labels inside participant/lane now show alignment snap lines
- Behavior is consistent with elements outside of participant
- Horizontal center alignment works correctly

Fixes #2371

### DIFF
--- a/lib/features/snapping/BpmnCreateMoveSnapping.js
+++ b/lib/features/snapping/BpmnCreateMoveSnapping.js
@@ -126,6 +126,17 @@ BpmnCreateMoveSnapping.prototype.addSnapTargetPoints = function(snapPoints, shap
 
   var snapTargets = this.getSnapTargets(shape, target);
 
+  // If shape is an external label, add its labelTarget as a snap target
+  // This ensures alignment works inside participant/lane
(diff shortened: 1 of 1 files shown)


ev-codes (changes requested):
@qianchongyang thanks for your contribution!

Before I continue the review, may you please:
- sign the CLA agreement [discussed here](https://github.com/bpmn-io/bpmn-js/pull/2402#issuecomment-4104335498)
- add tests that cover your change

Thank you 🙂

PR #2406, opened 2026-03-27
fix(behavior): remove invalid connections when replacing with event-based gateway

Zia-Ur-Rehman1:
## Summary
- When replacing an inclusive/exclusive gateway with an event-based gateway, outgoing sequence flows to non-event-based targets (regular Tasks) were not removed, violating BPMN rules
- Added logic in `EventBasedGatewayBehavior` to remove outgoing sequence flows to non-event-based targets during `shape.replace`
- Added `isEventBasedTarget` helper to identify valid event-based gateway targets (ReceiveTask, IntermediateCatchEvent with message/timer/conditional/signal definitions)

## Test plan
- [x] Existing test: removes non-event-based incoming sequence flows of event-based targets
- [x] New test: removes outgoing sequence flows to non-event-based targets when replacing with event-based gateway
- [x] New test: removes outgoing sequence flows to event-based targets that have other non-event-based incoming
- [x] New test: keeps outgoing sequence flows when replacing event-based gateway with inclusive gateway
- [x] All 1827 existing tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nikku:
Thanks for your contribution. Could you elaborate what bug the PR fixes? I don't see it, i.e. when trying to reproduce on [demo.bpmn.io](https://demo.bpmn.io):

capture j4t8s4_optimized

PR #2408, opened 2026-03-31
feat: activate direct editing for new activities and events

### LINKED ISSUE #748: Activate direct editing on create for all flow nodes (Event, Gateway, ...)
__Feature Request__

* Always activate direct editing on create. This would simplify quickly sketching new processes with labels.

__Note__

Right now, direct editing is only activated for `TextAnnotation` and `Task`.

__Screen Cast with proposed Interaction__

direct-edit

ev-codes:
### Proposed Changes

closes #748 

Immediately open the name editor after creating any new activity (not just tasks) or any new event.

I excluded plain start and end events because:
- These typically do not have labels associated with them
- Plain start events in subprocesses (based on team feedback) almost never have labels. I believe we should treat plain start events consistently.

https://github.com/user-attachments/assets/d02d4d2e-2ec8-4908-808a-01343e0961f7

### Checklist

### DIFF
--- a/lib/features/label-editing/LabelEditingProvider.js
+++ b/lib/features/label-editing/LabelEditingProvider.js
@@ -149,8 +149,7 @@ export default function LabelEditingProvider(
 
   function activateDirectEdit(element, force) {
     if (force ||
-        isAny(element, [ 'bpmn:Task', 'bpmn:TextAnnotation', 'bpmn:Participant' ]) ||
-        isCollapsedSubProcess(element)) {
+        isAny(element, [ 'bpmn:Activity', 'bpmn:Event', 'bpmn:TextAnnotation', 'bpmn:Participant' ])) {
 
       directEditing.activate(element);
     }
--- a/lib/features/label-editing/LabelEditingProvider.js
+++ b/lib/features/label-editing/LabelEditingProvider.js
@@ -7,7 +7,8 @@ import {
 } from '../../util/LabelUtil';
 
 import {
-  is
+  is,
+  getBusinessObject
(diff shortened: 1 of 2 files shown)


barmac:
> Plain start events in subprocesses (based on team feedback) almost never have labels.

There is a reason for this: The subprocess start event marks the initial token placement but does not depict any actual event in the business process sense.

lib/features/label-editing/LabelEditingProvider.js
```diff
@@ -517,4 +518,11 @@ function isExpandedPool(element) {
 
 function isEmptyText(label) {
   return !label || !label.trim();
+}
+
+function isPlainStartOrEndEvent(element) {
+  var businessObject = getBusinessObject(element);
+
+  return isAny(element, [ 'bpmn:StartEvent', 'bpmn:EndEvent' ]) &&
+    !(businessObject.eventDefinitions && businessObject.eventDefinitions.length);
```
barmac, line 527:
```suggestion
    !(businessObject.get('eventDefinitions').length);
```

barmac: Let's use moddle getter as the return type of array is guaranteed.

barmac (approved):
Let's try this out

lib/features/label-editing/LabelEditingProvider.js
nikku: Woops - @ev-codes I did not see @barmac's [comment](https://github.com/bpmn-io/bpmn-js/pull/2408#discussion_r3051930903). Could you follow up on that?

PR #2409, opened 2026-04-07
Security: Unescaped attribute/value composition in XML-like string helper

tuanaiseo:
## Problem

`elementToString` concatenates `e.$type` and `e.id` directly into an XML-like string without escaping. If these values are attacker-controlled and later rendered in HTML logs/UI, this can enable injection (markup break-out/XSS) or log forging.

**Severity**: `low`
**File**: `lib/import/Util.js`

## Solution

Escape XML/HTML special characters for dynamic fields (`<`, `>`, `&`, `"`, `'`) before interpolation, or return a structured object and serialize safely at output boundaries.

## Changes

- `lib/import/Util.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/lib/import/Util.js
+++ b/lib/import/Util.js
@@ -3,5 +3,18 @@ export function elementToString(e) {
     return '<null>';
   }
 
-  return '<' + e.$type + (e.id ? ' id="' + e.id : '') + '" />';
+  return '<' + escapeAttr(e.$type) + (e.id ? ' id="' + escapeAttr(e.id) : '') + '" />';
+}
+
+function escapeAttr(value) {
+  return String(value).replace(/[&<>"']/g, function(match) {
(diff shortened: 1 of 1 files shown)


nikku:
If you look at [`elementToString` use](https://github.com/search?q=repo%3Abpmn-io%2Fbpmn-js%20elementToString&type=code) then you see it is never provided as output to HTML - the goal is to use it as part of a log string. `id` is user input, but validated for XML compliance - ruling out that special chars may break out of the `""` jail. I don't see how `$type` is attacker controlled - maybe you can help me out here. 

To accept this PR, we'd need a test that verifies what is fixed (that previously broke).

PR #2412, opened 2026-04-09
wip: example textLayouter adaptation

Skaiir:
Example only, do not merge.

### DIFF
--- a/lib/draw/TextRenderer.js
+++ b/lib/draw/TextRenderer.js
@@ -1,7 +1,5 @@
 import { assign } from 'min-dash';
 
-import TextUtil from 'diagram-js/lib/util/Text';
-
 var DEFAULT_FONT_SIZE = 12;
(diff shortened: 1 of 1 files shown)


lib/draw/TextRenderer.js
```diff
@@ -113,7 +108,7 @@ export default function TextRenderer(config) {
    * @return {SVGElement} rendered text
    */
   this.createText = function(text, options) {
-    return textUtil.createText(text, options || {});
+    return textLayouter.createText(text, assign({ style: defaultStyle }, options || {}));
```
nikku, line 111: `textUtil` merged styles in the past (https://github.com/bpmn-io/diagram-js/pull/1024/changes#diff-e54070890a12bed25deb8aeaa067141398433466393c33dabc71a050a3dccf15R351) - this is no longer the case in this implementation.

PR #2437, opened 2026-05-30
fix(rules): disallow move when dropped outside the canvas

### LINKED ISSUE #2210: Moving elements beyond the border of a pool is possible when moving it outside the viewport

rkdfx:
### Proposed Changes

Closes #2210

Dragging an element off the canvas yields a `null` move target, which `BpmnRules#canMove` treated like move-start (`!target`) and allowed — bypassing move rules (e.g. stacking pools, moving a shape out of its pool). Now only an `undefined` target starts a move; a `null` target is rejected.

Builds on the fix from #2244 by @Adam-Thometz (added co-author); adds a rules-level test that fails without the change.

**Try it:** two pools, scroll one mostly off-screen, drag the other past the
  viewport onto it → the drop is now rejected.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to [bpmn-js](https://github.com/bpmn-io/bpmn-js) are documen
 
 ___Note:__ Yet to be released changes appear here._
 
+* `FIX`: disallow `move` when dropped outside of canvas ([#2437](https://github.com/bpmn-io/bpmn-js/pull/2437), [#2210](https://github.com/bpmn-io/bpmn-js/issues/2210))
 * `CHORE`: derive label link offset from (circular) outline shape ([#2457](https://github.com/bpmn-io/bpmn-js/pull/2457))
 
 ## 18.19.0
--- a/lib/features/modeling/behavior/FixHoverBehavior.js
+++ b/lib/features/modeling/behavior/FixHoverBehavior.js
@@ -45,8 +45,9 @@ export default function FixHoverBehavior(elementRegistry, eventBus, canvas) {
     var rootElement = canvas.getRootElement();
 
     // ensure bpmn:Group and label elements are dropped
-    // always onto the root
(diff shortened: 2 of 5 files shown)


barmac:
Hi @rkdfx,

Thanks for your contribution. We need you to sign the CLA before it can be reviewed and/or merged.

rkdfx:
> Hi @rkdfx,
> 
> Thanks for your contribution. We need you to sign the CLA before it can be reviewed and/or merged.

Hi, I already have signed it, it was not updated I guess. I updated it now. Thanks!

nikku (approved):
Thanks for your contribution @rkdfx - the change missed `Group` and `label` rules - for which the same behavior shall apply. Deeper investigation surfaced another bug, which I fixed via https://github.com/bpmn-io/bpmn-js/pull/2437/commits/3aee6932f49d2e6b56c39b78059b5ab327ebc888.

PR #2455, opened 2026-06-28
fix(snapping): snap labels to peers inside participant/lane

### LINKED ISSUE #2371: Alignment tools for labels don't work inside of participant/lane

eduardokaique:
### Proposed Changes

Closes #2371

#### Problem

External labels are always hovered onto the root element (see `FixHoverBehavior`),
so snap targets were computed from the root's children only. Inside a collaboration
the actual peers (events, gateways, ...) are children of the participant, not the
root — so no alignment assistance was given when moving a label inside a
participant/lane. Outside of a participant it worked because the label owner's
parent already *is* the root.

#### Solution

In `BpmnCreateMoveSnapping#addSnapTargetPoints`, when moving a label, also add the
label owner's actual container (`shape.labelTarget.parent`) as a snap target source,
reusing the existing recursion already used for sequence-flow parents. This restores
consistent behavior independent of the modeling context and also covers labels inside
expanded sub-processes.

### Steps to try out

1. Open `demo.bpmn.io` (or this build).
2. Create a pool with lanes and two labeled elements (e.g. start event + gateway).
3. Drag one external label — snap lines now align with the peer element inside the
   participant, identical to the behavior outside of a participant.

### Tests

Added `describe('label snapping')` in `BpmnCreateMoveSnappingSpec`:

- `should snap label to peer shape inside participant`
- `should snap label to peer shape (same behavior independent of context)` (regression in a plain process)

The participant test fails without the fix and passes with it; the full suite is green.

### Notes

This addresses the same issue as #2376 with a smaller, more general change (no extra
helper, also covers expanded sub-processes).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/features/snapping/BpmnCreateMoveSnapping.js
+++ b/lib/features/snapping/BpmnCreateMoveSnapping.js
@@ -171,6 +171,17 @@ BpmnCreateMoveSnapping.prototype.addSnapTargetPoints = function(snapPoints, shap
     snapPoints = this.addSnapTargetPoints(snapPoints, shape, target.parent);
   }
 
+  // labels are always hovered onto the root element (see FixHoverBehavior),
+  // so their peers (siblings of the labelTarget) are not picked up as snap
+  // targets; add the labelTarget's actual container explicitly
+  if (shape.labelTarget) {
+    var labelParent = shape.labelTarget.parent;
+
+    if (labelParent && labelParent !== target) {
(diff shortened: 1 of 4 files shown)


eduardokaique:
@falko, @nikku, @boris-petrov 
can you review this PR?

PR #2463, opened 2026-07-13
feat(popup-menu): add shared popup entries list

AlekseyManetov:
Related to https://github.com/camunda/camunda-modeler/issues/6037

### Proposed Changes

Adds `PopupEntries` — a shared source of BPMN element popup entry definitions (label, icon, target) exported from the module to be than reused in create/append/replace menus

Also make Replace menu width configurable via css variable to than make it wider to "360px" for better fit the descriptions(introduced in https://github.com/camunda/camunda-bpmn-js/pull/485) for most of elements.

Prerequisite for reusing the same definitions in the create/append menus — see https://github.com/bpmn-io/bpmn-js-create-append-anything/pull/88 for the full context.

### Checklist

### DIFF
--- a/lib/features/context-pad/ContextPadProvider.js
+++ b/lib/features/context-pad/ContextPadProvider.js
@@ -466,7 +466,7 @@ ContextPadProvider.prototype.getContextPadEntries = function(element) {
 
             popupMenu.open(element, 'bpmn-replace', position, {
               title: translate('Change element'),
-              width: 300,
+              width: 'var(--bpmn-replace-popup-width, 300px)',
               search: true
             });
           }
--- a/lib/features/popup-menu/PopupEntries.js
+++ b/lib/features/popup-menu/PopupEntries.js
@@ -1,0 +1,627 @@
+/**
+ * Shared catalog of BPMN element popup entry definitions.
+ *
+ * Each entry is the context-independent identity of an element — its `target`
+ * (moddle type + discriminating attributes), plus the `label` and `className`
+ * shared across every menu that offers it (create, append, replace).
+ *
+ * Menus compose their own option lists from these atoms, adding menu-specific fields
+ *
+ * @typedef { {
+ *   label: string;
+ *   className: string;
+ *   target: {
+ *     type: string;
+ *     isExpanded?: boolean;
+ *     triggeredByEvent?: boolean;
+ *     instantiate?: boolean;
+ *     eventGatewayType?: string;
+ *     isInterrupting?: boolean;
+ *     cancelActivity?: boolean;
+ *     eventDefinitionType?: string;
+ *     eventDefinitionAttrs?: Record<string, any>;
+ *   };
+ * } } ElementDefinition
+ *
+ * @type {Record<string, ElementDefinition>}
+ */
+export const PopupEntries = {
+
+  // Tasks
+  'task': {
+    label: 'Task',
+    className: 'bpmn-icon-task',
+    target: {
+      type: 'bpmn:Task'
+    }
+  },
+  'user-task': {
+    label: 'User task',
+    className: 'bpmn-icon-user',
+    target: {
+      type: 'bpmn:UserTask'
+    }
+  },
+  'service-task': {
+    label: 'Service task',
+    className: 'bpmn-icon-service',
+    target: {
+      type: 'bpmn:ServiceTask'
+    }
+  },
+  'send-task': {
+    label: 'Send task',
+    className: 'bpmn-icon-send',
+    target: {
+      type: 'bpmn:SendTask'
+    }
+  },
+  'receive-task': {
+    label: 'Receive task',
+    className: 'bpmn-icon-receive',
+    target: {
+      type: 'bpmn:ReceiveTask'
+    }
+  },
+  'manual-task': {
+    label: 'Manual task',
+    className: 'bpmn-icon-manual',
+    target: {
+      type: 'bpmn:ManualTask'
+    }
+  },
+  'rule-task': {
+    label: 'Business rule task',
+    className: 'bpmn-icon-business-rule',
+    target: {
+      type: 'bpmn:BusinessRuleTask'
+    }
+  },
+  'script-task': {
+    label: 'Script task',
+    className: 'bpmn-icon-script',
+    target: {
+      type: 'bpmn:ScriptTask'
+    }
+  },
+
+  // Sub-processes
+  'call-activity': {
+    label: 'Call activity',
+    className: 'bpmn-icon-call-activity',
+    target: {
+      type: 'bpmn:CallActivity'
+    }
+  },
+  'transaction': {
+    label: 'Transaction',
+    className: 'bpmn-icon-transaction',
+    target: {
+      type: 'bpmn:Transaction',
+      isExpanded: true
+    }
+  },
+  'event-subprocess': {
+    label: 'Event sub-process',
+    className: 'bpmn-icon-event-subprocess-expanded',
+    target: {
+      type: 'bpmn:SubProcess',
+      triggeredByEvent: true,
+      isExpanded: true
+    }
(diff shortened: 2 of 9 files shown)


lib/features/replace/ReplaceOptions.js
```diff
@@ -985,221 +264,46 @@ export var PARTICIPANT = [
...
-    actionName: 'replace-with-collapsed-pool',
-
-    // TODO(@janstuemmel): maybe design new icon
-    className: 'bpmn-icon-lane',
-    target: {
-      type: 'bpmn:Participant',
-      isExpanded: false
     }
-  }
+  })
```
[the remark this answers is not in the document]
AlekseyManetov: The (removes content) note is appended to the label only when the participant actually has children. The dynamic label already conveys the warning contextually, so duplicating it as an always-on description would be misleading

nikku (commented):
What is shipped in this PR is general description for _all_ elements. We'll not inject them "on the side", but ship them with the core experience. Is that what we want? If that is the case, then it is a good solution.

What we could / should (?) consider is whether we keep this optional (configurable) via an option => We ship, but folks can disable it.

Side-note: I'd be happier if we'd properly separate the two things happening here: 

* Popup Menu foundation 
* Adding description to all elements

AlekseyManetov:
> What is shipped in this PR is a general description for all elements. We'll not inject them "on the side", but ship them with the core experience. Is that what we want? 

Yes, we will inject them for the create, append, and replace menus. The replace functionality is shipped in this PR, while the create/append implementation will be a follow-up in `bpmn-js-create-append-anything`, where these menus live: https://github.com/bpmn-io/bpmn-js-create-append-anything/pull/88.

> What we could / should (?) consider is whether we keep this optional (configurable) via an option => We ship, but folks can disable it.

If there's a use case where someone needs to toggle description visibility, we can definitely add a configuration option (e.g., `popupMenu.open(..., { descriptions: false })`). However, I assumed everyone would benefit from these descriptions since they are generic to BPMN elements. I’m could to add the configuration if we see demand for it.

nikku:
> If there's a use case where someone needs to toggle description visibility, we can definitely add a configuration option (e.g., `popupMenu.open(..., { descriptions: false })`).

That would cancel _all_ descriptions - not only the new ones that now land in the core (when the core previously did not ship some). We'd need a different mechanism. The reason I'm raising this: If we build a new contract for other parts of the eco-system to rely upon, then we want to ensure that it works properly - including potential "hiding" of description, if users ask for it. Alternative is to plug-in the descriptions in our own tooling ("camunda-bpmn-js").

> However, I assumed everyone would benefit from these descriptions since they are generic to BPMN elements. 

Let's try that. It is a (UX) breaking change - but it is warranted because it improves the tooling.

AlekseyManetov:
> Side-note: I'd be happier if we'd properly separate the two things happening here

I've split this into two commits. 

@nikku Do you have any other suggestions or comments, or are we good to merge this?

barmac:
I've pushed https://github.com/bpmn-io/bpmn-js/pull/2463/commits/6bdd79e4c99c7cd97cea9183a03926f931fbcbe2 to verify that the descriptions can be overridden. So if a user wants to remove them, they only need to plug in a provider. No need for configuration.

lib/features/context-pad/ContextPadProvider.js
```diff
@@ -466,7 +466,7 @@ ContextPadProvider.prototype.getContextPadEntries = function(element) {
 
             popupMenu.open(element, 'bpmn-replace', position, {
               title: translate('Change element'),
-              width: 300,
+              width: 360,
```
barmac: Is this intended? The change is not related to the commit message.

lib/features/popup-menu/PopupEntries.js
```diff
@@ -31,55 +32,63 @@ export const PopupEntries = {
...
     className: 'bpmn-icon-user',
+    description: 'Work that needs to be done by a person',
     target: {
       type: 'bpmn:UserTask'
     }
   },
   'service-task': {
     label: 'Service task',
     className: 'bpmn-icon-service',
+    description: 'Automated work carried out by a job worker or connector',
```
barmac: This is Camunda-specific. It does not belong to bpmn-js.

lib/features/popup-menu/PopupEntries.js
```diff
@@ -31,55 +32,63 @@ export const PopupEntries = {
...
     target: {
       type: 'bpmn:ServiceTask'
     }
   },
   'send-task': {
     label: 'Send task',
     className: 'bpmn-icon-send',
+    description: 'Sends a message to an external participant',
```
barmac: "Participant" has a specific meaning in BPMN. Also a message does not to be passed to an external entity.

barmac:
@AlekseyManetov Some of the changes are Camunda-specific, e.g. [service task](https://github.com/bpmn-io/bpmn-js/pull/2463/changes/dd34e1a8e414cff4130cad19428cd82ede02cd68#r3622269777). IMO this is a good indicator that we'd rather implement descriptions in the camunda org. Alternative is that we keep BPMN-only approach, but that will mean that either Camunda Modeler users don't get good enough guidance, or we need to override the descriptions anyway.
My suggestion: Move the descriptions implementation to https://github.com/camunda/camunda-bpmn-js, and use a custom popup menu provider ("DescriptionProvider") like the one I added via the test.

AlekseyManetov:
> My suggestion: Move the descriptions implementation to https://github.com/camunda/camunda-bpmn-js, and use a custom popup menu provider ("DescriptionProvider") like the one I added via the test.

That makes sense and won't block us from providing Camunda-specific descriptions! I've reworked this PR and moved the description definition to `camunda-bpmn-js` (see https://github.com/camunda/camunda-bpmn-js/pull/485). 

This PR now only introduces a shared list of menu entry definitions, which we also need in order to define the create/append menus (see https://github.com/bpmn-io/bpmn-js-create-append-anything/pull/88). 

Additionally, it makes the replace menu wider (360px) to align it with the new width of create/append menu and better fit most element descriptions without wrapping them onto a second line. (There will still be a few long cases, but the wider menu reduces them significantly.)

@barmac Please take another look.

barmac:
IMO there's no good reason from the bpmn-js perspective to widen the replace menu.
image

If anything, I'd rather just drop the enforced width, and update diagram-js for a sensible behavior:

image

barmac:
Besides the width, I think it's OK to merge this.

AlekseyManetov:
> IMO there's no good reason from the bpmn-js perspective to widen the replace menu.

I've removed the width change from `bpmn-js` and instead made it configurable via provider config: https://github.com/bpmn-io/bpmn-js/pull/2463/changes/06eb61c918c262bffe0fb9fb148e503402081073. 

`camunda-bpmn-js` then adopts this: https://github.com/camunda/camunda-bpmn-js/pull/485/changes/473ab8f445978ba8df62bfbaae31eb7c3ba5d127.

lib/features/context-pad/ContextPadProvider.js
```diff
@@ -466,7 +466,7 @@ ContextPadProvider.prototype.getContextPadEntries = function(element) {
 
             popupMenu.open(element, 'bpmn-replace', position, {
               title: translate('Change element'),
-              width: 300,
+              width: 'var(--replace-popup-width, 300px)',
```
barmac: how about we use consistently `bpmn-replace`?

barmac:
```suggestion
              width: 'var(--bpmn-replace-popup-width, 300px)',
```

barmac (approved):
✅ with https://github.com/bpmn-io/bpmn-js/pull/2463/changes/47f9f7da5b0b1aa22d61abc91834ce4040828c36#r3638099450

PR #2465, opened 2026-07-15
feat: add tooltips with title + shortcut key for palette entries 

AlekseyManetov:
### Proposed Changes

Related to https://github.com/camunda/camunda-modeler/issues/6011
Depends on [bpmn-io/diagram-js#1082](https://github.com/bpmn-io/diagram-js/pull/1082)

Adds `shortcut` to the hand/lasso/space/global-connect tool entries so the diagram-js palette tooltip shows the key hint.

Additionally some refactoring of keyboard bindings registration and exposing `SHORTCUTS_KEY_BINDINGS` map.

https://github.com/user-attachments/assets/17baf956-6de9-4828-928b-27085f350909

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ All notable changes to [bpmn-js](https://github.com/bpmn-io/bpmn-js) are documen
 
 ___Note:__ Yet to be released changes appear here._
 
+* `FEAT`: add tooltip with title and shortcut on palette entries ([#2465](https://github.com/bpmn-io/bpmn-js/pull/2465))
 * `FIX`: point label link to the closest point of the connection ([#2493](https://github.com/bpmn-io/bpmn-js/pull/2493))
+* `DEPS`: update to `diagram-js@15.26.0`
 
 ## 18.26.0
 
--- a/lib/features/keyboard/BpmnKeyboardBindings.js
+++ b/lib/features/keyboard/BpmnKeyboardBindings.js
@@ -1,5 +1,7 @@
 import inherits from 'inherits-browser';
 
+import { forEach } from 'min-dash';
+
 import KeyboardBindings from 'diagram-js/lib/features/keyboard/KeyboardBindings';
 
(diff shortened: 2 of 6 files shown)


barmac:
[q] Why did you decide to implement it in bpmn-js? Note that: (1) palette is part of diagram-js; (2) we'd need to basically duplicate it for dmn-js

AlekseyManetov:
> [q] Why did you decide to implement it in bpmn-js?

Because the shortcut definitions themselves live in `bpmn-js`. But you are right that we would need to duplicate them for `dmn-js` in this case, which isn't a good solution. 

I would probably move the palette tooltip implementation to `diagram-js`. Then, `bpmn-js` and `dmn-js` can just pass the shortcut keys while defining palette entries. This way, the tooltip rendering logic will live in `diagram-js`, and `bpmn-js` / `dmn-js` will simply tell it which shortcut is used for each action in the palette. Does this sound better?

barmac:
The way I'd imagine this is that each palette provider can define a tooltip for an entry just like it defines now `title`. Then, that tooltip is used to display the actual tooltip.

AlekseyManetov:
I've reworked the solution and moved the palette tooltip implementation to `diagram-js` (see https://github.com/bpmn-io/diagram-js/pull/1082). Now, `bpmn-js` simply defines the shortcut key for palette entries, allowing it to be rendered inside the tooltip.

> The way I'd imagine this is that each palette provider can define a tooltip for an entry just like it defines now title.

I chose a slightly different approach by keeping `title` and `shortcut` as distinct properties. Because the shortcut is rendered inside a specific "badge" element, providing a single `tooltip` property would force us to pass a custom HTML structure into it to render that badge. 

Additionally, we don't want to keep the `title` property as a native HTML attribute anymore. If we do, the native browser tooltip will appear right alongside our custom one. 

Please take another look!

jarekdanielak (approved):

PR #2469, opened 2026-07-21
Add `categoryValueRef` support for elements visually inside groups

philippfromme:
### Proposed Changes

`categoryValueRef` was already used by `bpmn:Group` elements to reference the category value they represent. Other elements did not have that reference which made it impossible to tell what elements are part of a group without the DI. This pull request adds support for `categoryValueRef` for elements visually inside groups.

image

```xml
<bpmn:process id="Process_1" isExecutable="true">
  <bpmn:task id="Task_1" name="Task_1">
    <bpmn:categoryValueRef>CategoryValue_1</bpmn:categoryValueRef>
  </bpmn:task>
  <bpmn:task id="Task_2" name="Task_2" />
  <bpmn:group id="Group_1" categoryValueRef="CategoryValue_1" />
</bpmn:process>
<bpmn:category id="Category_1">
  <bpmn:categoryValue id="CategoryValue_1" value="Group_1" />
</bpmn:category>
```

* Add a reference to a group's category value to flow elements visually enclosed by it.
* Remove the reference when an element is moved out of the group or the group is removed.

### Checklist

* [ ] Pull request __establishes context__
  * [ ] __Link to related issue(s)__, i.e. `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`

### DIFF
--- a/lib/features/modeling/Modeling.js
+++ b/lib/features/modeling/Modeling.js
@@ -9,6 +9,7 @@ import AddLaneHandler from './cmd/AddLaneHandler';
 import SplitLaneHandler from './cmd/SplitLaneHandler';
 import ResizeLaneHandler from './cmd/ResizeLaneHandler';
 import UpdateFlowNodeRefsHandler from './cmd/UpdateFlowNodeRefsHandler';
+import UpdateCategoryValueRefsHandler from './cmd/UpdateCategoryValueRefsHandler';
 import IdClaimHandler from './cmd/IdClaimHandler';
 import SetColorHandler from './cmd/SetColorHandler';
 
@@ -37,6 +38,11 @@ import UpdateLabelHandler from '../label-editing/cmd/UpdateLabelHandler';
  * @typedef { {
  *   removeShape?: boolean;
  * } } UpdateLabelHints
+ *
+ * @typedef { {
+ *   categoryValue?: ModdleElement;
+ *   groupShape: Shape;
+ * } } GroupEntry
  */
 
 /**
@@ -87,6 +93,7 @@ Modeling.prototype.getHandlers = function() {
   handlers['lane.resize'] = ResizeLaneHandler;
   handlers['lane.split'] = SplitLaneHandler;
   handlers['lane.updateRefs'] = UpdateFlowNodeRefsHandler;
+  handlers['group.updateRefs'] = UpdateCategoryValueRefsHandler;
   handlers['id.updateClaim'] = IdClaimHandler;
   handlers['element.setColor'] = SetColorHandler;
   handlers['element.updateLabel'] = UpdateLabelHandler;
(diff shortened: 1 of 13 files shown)


nikku:
Great stuff :star: 

I'm a little bit more cautious about performance regressions these days, hence I gave the implementation a deeper look.

Turns out we can (should?) simplify it, so I pushed a few follow-up commits on top:

- **perf**: Set/Map-based dedup, hoisted category-value computation, scan flow elements only once per command bracket
- **refactor**: aligned `updateCategoryValueRefs` with the existing lane-refs API (plain shape arrays) and simplified the collector's `UpdateContext`
- **docs**: made the cleanup/revert contract explicit in `UpdateCategoryValueRefsHandler`
- **fix**: co-moving a group with a collapsed sub process stripped the ref, because the registry rescan picked up the sub process' root plane (shares the business object, has no position). Now filtered via `element.parent`, plus a regression test.

Verified update on co-move still recomputes once and yields zero net changes — bounded (O(elements) per bracket) and consistent with the lane-refs pattern, so I left it as-is.

nikku:
I'd be happy to fold my cat value changes into your commit. Just have a quick look into my adjustments @philippfromme.

philippfromme:
Thanks for the follow-ups, @nikku! Looks good to me. One decision I should make transparent: BPMN supports the inverse reference too: `categorizedFlowElements` points from the category to the flow elements. I decided not to implement that. 

###

image

###

Up to us to decide whether it's worth implementing that.

nikku (approved):
Approved. Needs @philippfromme's quick check + squash (that I'll execute).

nikku:
`categorizedFlowElements` - that is a virtual collection - we would need to update the bpmn model to implement it - let's not do that for now, but keep as a follow-up.

nikku:
One issue with the PR - caught during manual testing:

* :lady_beetle: When a group is created, then it does not have a `categoryValueRef` - hence there is nothing to categorize elements against

---

I take another stab at this.

philippfromme:
> 🐞 When a group is created, then it does not have a categoryValueRef - hence there is nothing to categorize elements against

But isn't this expected? Grouping will only really happen when you give the group a name. Or should we create an empty category value anyway?

nikku:
> Grouping will only really happen when you give the group a name. Or should we create an empty category value anyway?

We should make it possible to group elements with or without a group label - yes. I'm working on fixing that at the moment.

nikku:
Merged your cleanups already. Moving on to cleaning up the auto heal that I added.

nikku:
The following changes landed on this PR:

* categories now handled like other root elements
* eagerly create group -> category and update ("auto-heal") on group change
* actual fix for your matter (https://github.com/bpmn-io/bpmn-js/pull/2469/commits/b88f8c69bf3e06e3b6b33f9f3306d391ef5293b5)

Manual integration test passed.

@philippfromme please have a look.

nikku (approved):
Ready to merge from my end ✔️

nikku:
Released via https://github.com/bpmn-io/bpmn-js/releases/tag/v18.23.0.

PR #2475, opened 2026-07-27
fix(rules): prevent copying message flow without its participants

### LINKED ISSUE #1902: Creating Message Flow in Process Not Prevented

nikku:
### Proposed Changes

Copying a `bpmn:MessageFlow` together with its two flow nodes — but _not_ their participants — and pasting after the diagram became a plain process crashed the editor (`no parent for <MessageFlow_1> in <Process_...>`). The message flow slipped through the per-connection `connection.create` rule and landed in a `bpmn:Process`, where `BpmnUpdater` pushed onto an undefined `messageFlows` list.

This blocks the case at the `element.copy` rule chokepoint in `BpmnRules`: a message flow may now only be copied if **both** of its connected participants are part of the copied selection, mirroring the existing "copy lane without parent participant → false" precedent. When they aren't, the message flow is silently dropped from the clipboard instead of being pasted into an invalid organizational context:

capture Tk1fj7_optimized

Closes https://github.com/bpmn-io/bpmn-js/issues/1902

**Steps to try out**

1. Model a collaboration with two pools, a task in each, and a message flow between the tasks.
2. Copy both tasks + the message flow (without the pools).
3. Remove both pools so the diagram becomes a plain process.
4. Paste — previously crashed; now the tasks paste and the message flow is dropped.

### Checklist

### DIFF
--- a/lib/features/rules/BpmnRules.js
+++ b/lib/features/rules/BpmnRules.js
@@ -1190,7 +1190,45 @@ function canCopy(elements, element) {
     return false;
   }
 
+  // do not copy message flows unless both connected participants are copied,
+  // too; otherwise they would end up in an invalid organizational context
+  if (is(element, 'bpmn:MessageFlow') && !canCopyMessageFlow(elements, element)) {
+    return false;
+  }
+
   return true;
+}
+
+/**
+ * A message flow may only be copied if both of its participants are copied,
+ * too. Otherwise it would be pasted into an invalid organizational context,
+ * e.g. a plain process.
+ *
+ * @param {Element[]} elements
+ * @param {Connection} connection
+ *
+ * @return {boolean}
+ */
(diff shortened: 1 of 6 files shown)


nikku:
Released via https://github.com/bpmn-io/bpmn-js/releases/tag/v18.23.0.

PR #2476, opened 2026-07-28
fix(export): pad exported SVG to avoid clipping strokes

jarekdanielak:
### Proposed Changes

Exporting a freshly opened diagram produced a slightly smaller image with clipped edge strokes (e.g. start-/end-event borders), and the exported size depended on hover/selection state. Since [diagram-js#1064](https://github.com/bpmn-io/diagram-js/pull/1064) (15.19) outlines are created lazily, so they no longer implicitly pad the export bounds.

`saveSVG` now hides outlines while measuring the bounds and pads them by 5px, keeping exports stable and unclipped. Requires `diagram-js@^15.23.2` (ships the `.djs-outline-hidden` class).

Related to https://github.com/camunda/camunda-modeler/issues/6065

### Steps to try out

1. Open a diagram with an element flush against the right/bottom edge (e.g. an end event).
2. Export to SVG/PNG **without** hovering or selecting anything.
3. The full border is visible (before: clipped), and the size no longer changes based on prior interaction.

### Checklist

### DIFF
--- a/lib/BaseViewer.js
+++ b/lib/BaseViewer.js
@@ -14,6 +14,7 @@ import {
 import {
   domify,
   assignStyle,
+  classes as domClasses,
   query as domQuery,
   remove as domRemove
 } from 'min-dom';
@@ -30,6 +31,23 @@ import inherits from 'inherits-browser';
 import {
   importBpmnDiagram
 } from './import/Importer';
+
+/**
+ * Padding added around the diagram bounds on SVG export.
+ *
+ * `getBBox` returns the geometry bounds only, excluding element strokes, so
+ * without padding strokes at the diagram edges get clipped. Historically this
+ * padding was provided implicitly by element outlines, which diagram-js now
+ * creates lazily (diagram-js@15.19, bpmn-io/diagram-js#1064).
+ */
+const EXPORT_PADDING = 5;
+
+/**
+ * Class set on the canvas container to hide element outlines, which diagram-js
+ * excludes from rendering while present. Applied when measuring the export
+ * bounds so lazily created outlines do not skew them.
+ */
+const OUTLINE_HIDDEN_CLS = 'djs-outline-hidden';
 
 /**
  * @template T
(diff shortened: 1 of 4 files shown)


AlekseyManetov:
> dmn-js (dmn-js-drd) has the same getBBox() → viewBox pattern and is exposed to the same clipping. Happy to open a follow-up there.

That's right, we need the same fix for dmn-js too.

AlekseyManetov (approved):

nikku:
This does not seem to account for https://github.com/camunda/camunda-modeler/issues/6065#issuecomment-5104429427 - when outlines are present, then the padding is now + 10 pixel - so the export depends on previously selected elements. Consider if it is worth to fix this.

jarekdanielak:
@nikku in https://github.com/bpmn-io/bpmn-js/pull/2476/commits/33f9db804b5a5f2ee4f4d9ebc05e56976e83bb9f the outline layer is temporarily removed from DOM to take an accurate measurement.

WDYT?

Alternative considered: `getBBox(elementRegistry.getAll())` + extra padding. No DOM touching, but blind to customly rendered elements.

lib/BaseViewer.js
```diff
@@ -41,6 +42,42 @@ import {
  */
 const EXPORT_PADDING = 5;
 
+/**
```
nikku, line 45: What about instead temporarily setting them as `display: none` - this removes the elements from the render tree.

Can be supported through a flag that we set on the canvas root (temporarily).

nikku: We already ship custom styles, so hiding outlines is: Add a class to SVG -> export -> remove class from SVG.

jarekdanielak: See my proposal for toggling outlines in diagram-js: https://github.com/bpmn-io/diagram-js/pull/1094

jarekdanielak:
@nikku updated, using the new diagram-js flag.

nikku (approved):
Good change ⭐

PR #2497, opened 2026-09-10
feat: source variables from @bpmn-io/theme

AlekseyManetov:
Related to https://github.com/bpmn-io/internal-docs/issues/1364

### Proposed Changes

bpmn-js declared twenty-four private palette colours and read five of them. A theme had no semantic name to override — it had to pin every component variable by hand.

- The private palette is replaced by the semantic tokens of [`@bpmn-io/theme`](https://github.com/bpmn-io/shadcn-theme/tree/main/packages/theme) it actually uses, copied verbatim into the top of `bpmn-js.css` and declared on `.bio-theme-parent`. Nothing imports the theme at runtime — the copy *is* the shipped stylesheet.
- `BaseViewer` puts `bio-theme-parent` on the container, so the tokens are in scope.
- `npm run lint:theme`, part of `npm run all`, fails if that copy drifts from the package.
- `--button-border-radius` and `--surface-border-radius` bind the two corner radii to the `--bio-radius-*` scale, declared on `.bjs-container`.
- One dead variable is removed.

### Rendered output

Every `var()` chain was resolved in both revisions and the final colours compared. **All 10 component variables and all 21 property declarations are identical** — nothing renders differently.

`--breadcrumbs-arrow-color` is removed. The breadcrumb arrow is an SVG loaded through `content: url()`, which makes the pseudo-element replaced content — `color` cannot reach inside it, and the SVG has no `currentColor` fill. Forcing the variable to `red` renders a byte-identical image.

### Corner radii

Both radii were literal `2px` and are now bound to the shared scale, matching the element kinds used across the ecosystem:

| element | variable | was | now |
| --- | --- | --- | --- |
| `.bjs-breadcrumbs` | `--surface-border-radius` (`--bio-radius-lg`) | `2px` | `4px` |
| `.bjs-drilldown` | `--button-border-radius` (`--bio-radius-md`) | `2px` | `3px` |

### Steps to try out

```bash
npx @bpmn-io/sr bpmn-io/bpmn-js#shared-base-theme
```

### Checklist


nikku (approved):
Works ⭐

barmac (approved):

